Issue #13, opened 2017-07-12
Automate license header for all new files


swese44:
This will be easy to forget to add to all new files:

`/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */`

unindented:
Fixed by https://github.com/Microsoft/YamUI/pull/17

PR #17, opened 2017-07-13
Add eslint-plugin-header to check copyright notices

unindented:
Addresses part of https://github.com/Microsoft/YamUI/issues/13

We'd need to do something equivalent with `tslint`... I may look into it when I have time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "enzyme": "=2.8.2",
     "eslint": "=3.19.0",
     "eslint-config-airbnb": "=14.1.0",
+    "eslint-plugin-header": "=1.0.0",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
     "enzyme": "=2.8.2",
     "eslint": "=3.19.0",
     "eslint-config-airbnb": "=14.1.0",
+    "eslint-plugin-header": "^1.0.0",
```
swese44: `=`

swese44:
this is awesome thanks

unindented:
Ok, turns out `tslint` has a built-in rule for this, `file-header`. We're ready then!

PR #19, opened 2017-07-13
Update Fabric dependency

swese44:
having a dependency issue locally, want to rule out fabric after re-publishing


swese44:
@annr ya this has happened a few times during Fabric updates, i believe it's from their default props changing

PR #25, opened 2017-08-04
Button onHoverIn and onHoverOut events

swese44:


src/components/Button/button.test.tsx
```diff
@@ -163,22 +163,42 @@ describe('<Button />', () => {
...
+    function onHoverIn() {
+      hoveredIn = true;
+    }
+    function onHoverOut() {
+      hoveredOut = true;
+    }
     beforeEach(() => {
       clicked = false;  
+      hoveredIn = false;  
+      hoveredOut = false;
```
unindented: Why not use spies (`jest.fn()`) instead of functions that change a boolean?

src/components/Button/button.test.tsx
```diff
@@ -163,22 +163,42 @@ describe('<Button />', () => {
...
       component = shallow(
-        <Button text={sampleText} onClick={clickMe} />,
+        <Button text={sampleText} onClick={onClick} onHoverIn={onHoverIn} onHoverOut={onHoverOut} />,
```
benlakey: Consider batching up the props in an object and then using spread to make it more concise as props are added:

```javascript
<Button {...props} />
```

unindented: But then you don't really know which props your forwarding. I'm more comfortable being explicit here.

src/components/Button/button.test.tsx
```diff
@@ -163,22 +163,42 @@ describe('<Button />', () => {
...
       );
     });
 
     it('triggers its onClick callback', () => {
       component.simulate('click');
       expect(clicked).toBe(true);
```
benlakey: In combination with the feedback from @unindented  (spies), recommend leveraging `describe` to describe scenarios and `it` to specify expectations:

```javascript
describe('when the component is clicked', function() {
  beforeEach(function() {
    component.simulate('click');
  });

  it('triggers the onClick callback', function() {
    expect(onClick).toHaveBeenCalled();
  });
});

Comment (benlakey) [inline src/components/Button/button.test.tsx]: same comments for other tests below

File: src/components/Button/button.test.tsx (hunk 1 above)
Comment (swese44) [inline src/components/Button/button.test.tsx]: fixed thanks

File: src/components/Button/button.test.tsx (hunk 3 above)
Comment (swese44) [inline src/components/Button/button.test.tsx]: added thanks

File: src/components/Button/button.test.tsx (hunk 2 above)
Comment (benlakey) [inline src/components/Button/button.test.tsx]: Its typescript?

**Comment by annr** [comment]
We should probably update the README for Button as well.

File: src/components/Button/button.test.tsx (hunk 2 above)
Comment (unindented) [inline src/components/Button/button.test.tsx]: Ha, you're right.

File: src/components/Button/button.test.tsx [hunk 4]
```diff
@@ -163,21 +163,50 @@ describe('<Button />', () => {
...
+      });
+    });
+
+    describe('when hovered out', () => {
+      beforeEach(() => {
+        component.simulate('mouseLeave');
+      });
+
+      it('triggers its onHoverOut callback', () => {
+        expect(onHoverIn).not.toHaveBeenCalled();
```
annr: Testing onHoverIn/onHoverOut "not called" is not necessary. We're using onMouseLeave/onMouseEnter on the component, and you're testing what those native attributes do. Since one handler is passed, it isn't a valuable test for us.

swese44: cool removed

swese44:
* Changed onHoverIn and onHoverOut to onMouseEnter and onMouseLeave, this should be more intuitive with the standard event names
* Added onFocus and onBlur so TriState buttons can respond to keyboard events
* Added font-family to button (it was defaulting to the browser's default font-family - we can remove $fontStack once we wrap our app in a Fabric component
* Updated React peer-dependency to be flexible so we don't block YamJS (or possibly SharepointWebParts) from updating its version of React

swese44:
@annr could you review again? also you should be able to set your github creds just for a single repo

PR #26, opened 2017-08-08
LinkButton => href prop for Button

swese44:
- Adds `href` to Button, which renders a link styled as a button
- Disallows `disabled` if `href` is provided
- Adds custom PropTypes validation to complain about bad href values like `#` and `javascript://`


src/components/Button/button.css
```diff
@@ -13,6 +13,10 @@
   border-width: 1px;
   border-style: solid;
   box-shadow: 0 1px 1px 0 rgba(0,0,0,.05);
+}
+
+.y-button,
+.y-button:hover {
```
unindented: `:focus` pseudoselector doesn't apply here?

swese44: nope, this was strictly to remove a link's underline on hover (not focus)

src/components/Button/index.tsx
```diff
@@ -71,17 +74,63 @@ export interface ButtonProps extends BaseComponentProps {
...
+    const { ariaLabel, text, icon, onClick, onMouseEnter, onMouseLeave, onFocus, onBlur } = props;
+
+    // Only use href for LinkButton, and only use disabled for non-LinkButton
+    let disabled;
+    let href;
+    if ((props as RegularButtonProps).disabled) {
+      disabled = (props as RegularButtonProps).disabled;
+    } else if ((props as LinkButtonProps).href) {
+      href = (props as LinkButtonProps).href;
+    }
```
unindented: I suspect @benlakey would frown upon this bit of local state. Is there a way we can express this with `const` instead of `let`? I'm totally fine with a ternary operator, but I'm open to other ideas.

swese44: i'd argue it's not really state, more like massaging incoming props. do you have a specific suggestion on how to clean this up?

unindented: Right, not state in terms of React's state, but local state to the function. @benlakey always tries to convert `let` to `const`, cause that simplifies reasoning about those variables, and I've gotten into the habit too.

Would this be equivalent?

```js
const disabled = (props as RegularButtonProps).disabled || undefined;
const href = !disabled && (props as LinkButtonProps).href || undefined;
```

benlakey: Right. Its about mutability. When you use `let` you open yourself up to the possibility that your local state will get mutated and mutation is the cause of a lot of bugs. With `const` you can be sure that assignment was only ever used once, which means you do not need to dig through the code to see where a thing was mutated; its starting value is its current value.

benlakey: It also allow compilers to perform optimizations, but thats less interesting than the human side of it.

swese44: i'll remove the if/else and use const. the if/else should not be necessary now that the typescript interface xors href and disabled for us

swese44: @unindented is the `|| undefined` necessary?

swese44: updated thanks

unindented: I was just trying to keep the same behavior as the original, where, if you didn't go into the respective branch, the value would be undefined.

If that's unnecessary for your logic, feel free to remove.

PR #27, opened 2017-08-09
HorizontalList components

swese44:
Simple HorizontalList and ListItem components. In the future I think we might move ListItem out of the HorizontalList directory and be shared by more components, but not necessary right now.


src/components/HorizontalList/horizontal-list.css
```diff
@@ -0,0 +1,20 @@
...
+  flex-flow: row wrap;
+}
+
+.y-horizontalList__align-right {
+  justify-content: flex-end;
+}
+
+.y-horizontalList--item {
+  padding: 0;
+  margin: 0 $gutterWidth__xSmall 0 0;
```
annr: Why are we importing gutterWidth?

src/components/HorizontalList/horizontal-list.tsx
```diff
@@ -0,0 +1,30 @@
...
+    classes.push('y-horizontalList__align-right');
+  }
+  if (props.className) {
+    classes.push(props.className);
+  }
+  return classes.join(' ');
+};
+
+export const HorizontalList: React.StatelessComponent<HorizontalListProps> = (props) =>
+  <ul className={getClasses(props)}>{props.children}</ul>;
```
annr: I wonder if we'll need to support an ordered list as well as an unordered list. I thought the list of members ("facepile") in hovercards was an ordered list, but I just checked and we didn't implement it that way (it's ordered by featured rating or whatever)

src/components/HorizontalList/horizontal-list.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    it('matches its snapshot', () => {
+      expect(component).toMatchSnapshot();
+    });
+  });
+
+  describe('with child content', () => {
+    beforeEach(() => {
+      component = shallow(
+        <HorizontalList>list content</HorizontalList>,
+      );
```
annr: I don't know if this bothers you, but this is invalid HTML. `<ul>` must have `<li>` children. It's just a test tho.

swese44: i thought about testing both components together but it didn't seem like a good "unit" test. between this and visual diffs i think we're covered

src/components/HorizontalList/horizontal-list.tsx
swese44: hmm hadn't thought of that. we can always add it if/when we need it.

@BrendanMcK

src/components/HorizontalList/horizontal-list.css
swese44: for consistency with our vertical rhythm, if it changes we can change it in just one place. also gutters are currently specified in `rem`; say we decide gutters should remain fixed when a user changes their browser's base font size then we could change them to `px` in one place rather than throughout the codebase in every layout component

PR #28, opened 2017-08-10
Upgrading Fabric, TypeScript and other necessary dependencies

swese44:
Upgrading YamUI to TypeScript 2.4, moving its dependency to peerDependency as YamJS now depends on TS directly.

Fixed a Fabric upgrade issue in Avatar with more specific CSS (thanks visual diff test!). Also fixed linting errors from new linting rules.


unindented:
Did you encounter any issues? This looks good to me.

package.json
```diff
@@ -54,19 +54,20 @@
...
-    "typescript": "=2.3.4"
+    "tslint": "=5.6.0",
+    "tslint-config-airbnb": "=5.2.1",
+    "typescript": "=2.4.2"
   },
   "peerDependencies": {
     "react": "^15.0.0",
-    "react-dom": "^15.0.0"
+    "react-dom": "^15.0.0",
+    "typescript": "^2.4.2"
```
unindented: Could we be a bit more lax here, with `^2.4.0` maybe?

swese44:
@unindented no major issues, mainly just figuring out which other modules needed to also get updated

PR #29, opened 2017-08-17
Button tweaks

annr:


unindented:
Ah sorry, didn't see the `wip` label! Ignore my approval.

annr:
I love the approval!

I have new visual diffs, and support for icons on the right, but I'm going to try to get sign-off from Adrian before I commit all that. Although maybe I should check in the new code for approval...

src/components/Button/index.tsx
```diff
@@ -20,6 +20,12 @@ export const ButtonSize = Enum({
 });
 export type ButtonSize = Enum<typeof ButtonSize>;
 
+export const IconPosition = Enum({
```
swese44: Do you want to use string literal types for simple options like `right|left`? We have them in a few other places:
https://github.com/Microsoft/YamUI/blob/master/src/components/Block/index.tsx#L35
https://github.com/Microsoft/YamUI/blob/master/src/components/HorizontalList/horizontal-list.tsx#L9

annr: Consistency is good, and I think I like this better anyway. I'l make this change after I get CI to pass.

PR #30, opened 2017-08-24
Convert SVG icons to React

swese44:
- Renders an inline SVG rather than linked in a `<use>` element
- Removes the need for a loader for Storybook (or consumers like YamJS)
- Removes the need for Jest file mock
- [ ] Get updated SVGs for `external`, `lock`, `no` and `time` icons normalized to 320x320


unindented:
I fear some illustrations may have more than a single path, so this approach may be too limiting.

Given that React supports the full SVG spec, and there are tools like https://www.npmjs.com/package/svg-to-react-cli, why not convert the SVG files that designers give us into React components, and then have `<Icon>` render the right one as its child?

src/components/Icon/paths.ts
```diff
@@ -0,0 +1,25 @@
...
+  phone: 'M266.654 289.565c-68.082 4.005-125.864-19.655-171.26-70.15C50.98 170.01 28.917 114.632 30.04 54.298c.25-13.434 11.17-24.17 24.536-24.3h51.908c4.48 0 8.405 3 9.574 7.32l15.252 56.36c1.43 5.28-1.698 8.717-6.986 12.144-.01.003-9.134 5.79-27.37 17.36 8.53 24.79 13.927 41.446 34.336 64.15 20.99 23.347 37.747 30.055 64.71 38.296l18.063-29.926c3.44-5.327 6.886-8.45 12.173-7.024l56.435 15.232c4.33 1.168 7.33 5.087 7.33 9.56v51.84c-.24 12.813-10.32 23.487-23.342 24.253z',
+  plus: 'M140.025 140H40.5c-11.046 0-20 8.954-20 20s8.954 20 20 20h99.525v99.525c0 11.045 8.954 20 20 20 11.045 0 20-8.955 20-20V180h99.524c11.046 0 20-8.954 20-20s-8.954-20-20-20h-99.524V40.475c0-11.045-8.955-20-20-20-11.046 0-20 8.955-20 20V140z',
+  // TODO: time is 280x280, needs to be normalized
+  time: 'M0 140c0 77.32 62.68 140 140 140s140-62.68 140-140S217.32 0 140 0 0 62.68 0 140zm240 0c0 55.228-44.772 100-100 100S40 195.228 40 140 84.772 40 140 40s100 44.772 100 100zm-90-56c0-7.732-6.268-14-14-14s-14 6.268-14 14v52c0 3.797 1.512 7.242 3.966 9.764.906.953 1.962 1.793 3.158 2.484l51.962 30c6.696 3.866 15.258 1.572 19.124-5.124 3.866-6.696 1.572-15.258-5.124-19.124L150 127.97V84z',
+};
```
annr: This is called paths but it's actually inline data.

annr:
That could be really clean, and no harder to add new Icons since we have to run a tool on them anyway.

src/components/Icon/paths.ts
swese44: oh good call

swese44:
we're just talking about single-color icons here, not full illustrations. i'll meet with @mamuso at 2 and we'll see if there's a possibility of an icon having more than 1 path, but i think @unindented's suggestion may have more benefits.

we could use `svg-to-react-cli` if we chain it behind another tool to strip out the elements and attributes we don't want, maybe `svgo`. i'll look at that in a separate PR to make sure we can strip all the extra gunk we don't want.

PR #32, opened 2017-08-26
Compile to es5 for consumers which don't run loaders for node_modules

swese44:


unindented:
Components are imported by doing something like `import Button from 'yamui/dist/components/Button'` right? So it's not like by going with ES5 we'll be stopping webpack from tree-shaking (because we weren't relying on tree-shaking in the first place).

swese44:
@unindented Correct

PR #34, opened 2017-08-29
Convert SVGs to React components

annr:
A reminder of what this branch does:

 - removes svg loaders
 - removes svgMock file
 - suppresses test coverage requirements for React SVGs
 - moves raw and prepared SVGs into /assets/Icons folder which has a README for SVG prep
  (there's a skype icon in there that is not used yet)
 - adds all the SVGs as react components
 - updates diffs


src/components/Icon/index.tsx
```diff
@@ -45,12 +44,12 @@ export default class Icon extends React.Component<IconProps, {}> {
 
   public render () {
     const { size, icon } = this.props;
-
-    return (
-      <svg className={this.getClasses()} style={this.getInlineStyles()} aria-hidden="true">
-        <use xlinkHref={`#${icon}`} />
-      </svg>
-    );
+    return React.createElement(ICONS[icon], {
```
swese44: Do we need `React.createElement` here, or could we do something like this?

```
const CurrentIcon = ICONS[icon];
return <CurrentIcon className={} style={} />
```

src/components/Icon/icons/Phone.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export default function Phone(props: any) {
```
swese44: is this an example Icon to copy+paste from?

src/components/Icon/icons/Attach.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export default function Attach(props: any) {
+  return (
+    <svg viewBox="0 0 320 320" xmlns="http://www.w3.org/2000/svg" xmlnsXlink="http://www.w3.org/1999/xlink" aria-hidden="true" {...props}>
```
swese44: these might need height and width attributes to match the viewBox bounds for proper scaling, both 320 in this case (although some of our icons have different dimensions right now)

i think we should also delete the `xmlnsXlink` attributes since we're not using `<anchor>` links in SVG:
https://en.wikipedia.org/wiki/XLink

src/components/Icon/icons/Check.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export default function Check(props: any) {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" xmlnsXlink="http://www.w3.org/1999/xlink" viewBox="0 0 320 320" aria-hidden="true" {...props}>
+      <path id="a" d="M56.142 166.657c-7.81-7.816-20.474-7.816-28.284 0-7.81 7.815-7.81 20.487 0 28.303l68.13 68.178c7.812 7.816 20.475 7.816 28.285 0l151.87-151.973c7.81-7.815 7.81-20.487 0-28.303-7.81-7.816-20.475-7.816-28.285 0L110.13 220.682l-53.988-54.025z" />
```
swese44: should delete the id in this path

src/components/Icon/icons/Group.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export default function Group(props: any) {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" xmlnsXlink="http://www.w3.org/1999/xlink" aria-hidden="true" {...props}>
+      <path id="a" d="M212 146c-23.196 0-42-18.804-42-42s18.804-42 42-42 42 18.804 42 42-18.804 42-42 42zm-104 0c-23.196 0-42-18.804-42-42s18.804-42 42-42 42 18.804 42 42-18.804 42-42 42zm104 20c18.002 0 36.004 1.367 54.005 4.1 19.55 2.97 33.995 19.775 33.995 39.547V258h-84v-48.353c0-16.9-7.037-32.357-18.517-43.35 4.84-.198 9.678-.297 14.517-.297zm-104 0c18.002 0 36.004 1.367 54.005 4.1 19.55 2.97 33.995 19.775 33.995 39.547V258H20v-48.353c0-19.772 14.446-36.578 33.995-39.546 18-2.73 36.003-4.1 54.005-4.1z" />
```
swese44: looks like there are several ids we need to delete

package-lock.json
```diff
@@ -0,0 +1,9817 @@
+{
+  "name": "yamui",
```
swese44: did you switch to yarn to npm? do we want to manage both yarn.lock and package-lock.json?

src/components/Icon/icons/Lock.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export default function Lock(props: any) {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" {...props}>
```
swese44: this needs a `viewBox` attribute. note that Lock's height/width did not match the 320x320 most other icons had. you'll probably need to make a pass to make sure all React icons' viewBox/height/width attributes match their original SVG

swese44:
are you thinking we could hand-craft these React icons in the near-term to unblock a release, then add CLI tasks later?

src/components/Icon/icons/Phone.tsx
annr: I cannibalized what came out of the svgtoreact cli.

annr: We would make a proper template, and ideally create our own command line tool if we go in this direction.

package-lock.json
annr: Sorry, no, I shouldn't have checked this in. This exploratory branch is a mess! Cleaning up now and will be ready for review this am.

src/components/Icon/index.tsx
annr: I thought I had tried that! But now it works.

src/components/Icon/icons/Attach.tsx
annr: Agreed. 

I think svgo is going to remove xmlnsXlink, so we don't have to worry about that, unless for some reason we can't use svgo (it may be messing up some of the icons). But we do need to make sure that height and width are set as well as viewBox (our svgs don't always have viewBox).

src/components/Icon/icons/Check.tsx
annr: It's gone. I don't think I manually removed it -- I think SVGO did.

annr:
The cli we want to run is svgo, which optimizes the SVG in a number of ways. svg-to-react cli didn't work for us, or maybe we can fix it to work for us. I made a little shell script to create our React SVGs from a template, but we'll want it all together in one tidy task for ourselves, I think. I can do that later.

assets/Icons/optimized/attach.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="320" height="320" viewBox="0 0 320 320" xmlns="http://www.w3.org/2000/svg">
```
swese44: are these the new icons from Manuel? if so these should all use 240 instead of 320

assets/Icons/optimized/mail.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="320" height="320"><path d="M20.647 88.803C23.317 76.896 33.954 68 46.667 68h226.666c12.657 0 23.255 8.818 25.984 20.646l-139.47 70.977-139.2-70.82zM20 121.22l132.435 64.915c2.334 1.348 4.907 1.91 7.41 1.767 2.505.142 5.077-.42 7.41-1.767L300 121.042v120.29c0 14.73-11.94 26.668-26.667 26.668H46.667C31.94 268 20 256.06 20 241.333V121.22z"/></svg>
```
swese44: missing viewBox

assets/Icons/optimized/phone.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="320" height="320"><path d="M266.654 289.565c-68.082 4.005-125.864-19.655-171.26-70.15C50.98 170.01 28.917 114.632 30.04 54.298c.25-13.434 11.17-24.17 24.536-24.3h51.908c4.48 0 8.405 3 9.574 7.32l15.252 56.36c1.43 5.28-1.698 8.717-6.986 12.144-.01.003-9.134 5.79-27.37 17.36 8.53 24.79 13.927 41.446 34.336 64.15 20.99 23.347 37.747 30.055 64.71 38.296l18.063-29.926c3.44-5.327 6.886-8.45 12.173-7.024l56.435 15.232c4.33 1.168 7.33 5.087 7.33 9.56v51.84c-.24 12.813-10.32 23.487-23.342 24.253z"/></svg>
```
swese44: missing viewBox

assets/Icons/optimized/mail.svg
swese44: oh duh this is in Assets nevermind

assets/Icons/optimized/attach.svg
swese44: seems like these icons are out of date?

assets/Icons/raw/skype.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
swese44: need to remove:
* xml tag
* comments
* title, desc, defs tags
* ids from g tags (4 of them)

swese44: wait this is in the `raw` folder. why do we have icons in `Assets` and `components/Icon/raw` ?

annr: As a reference in case we need to re-optimize the SVG.

annr: I haven't received the 240v240 icons yet. Hopefully we can develop a process where UX keeps the assets/Icons folder updated. I'll talk to him today.

There is no reason why we can't go out with the currently-sized icons which are the ones we have right now.

annr: If the raw folder causes confusion, let's get rid of it. I have found it important to be able to reference the original SVG, however.

assets/Icons/optimized/attach.svg
annr: I mentioned this below but we don't have the new ones yet.

assets/Icons/optimized/mail.svg
annr: Still it should be correct here. Basically we have replace all these and make sure they are right when we get the new ones.

assets/Icons/raw/skype.svg
swese44: is "raw" the same as what Design provides in the Assets directory?

annr: The assets/Icons folder did not exist before. I gave that folder the name `raw`. Manuel can choose another label if he'd like. I'll talk to him soon.

swese44: ok i see now. the optimized version should be a temporary "build" step once we have this task automated, but let's push this out and worry about that later

annr:
Okay, I'm going to merge this and if Manuel wants to change the process we can do that today.

PR #38, opened 2017-09-05
Fixing visual diffs for the html baseline page

swese44:


swese44:
i think there are multiple points of failure in the fake markup page. i'm abandoning this but we can use this branch as a reference later if necessary

PR #39, opened 2017-09-07
Remove 16.3KB of extra Fabric Button types we don't use

swese44:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/travisci/travis-node-js:latest
+FROM travisci/ci-nodejs:packer-1494866191
```
unindented: Where's this version number coming from? Is it just so that we point to a specific version, and don't get bitten by random updates?

.travis.yml
```diff
@@ -1,10 +1,12 @@
+sudo: false
+dist: trusty
 language: node_js
 
 node_js:
-  - "6"
+  - "6.10.3"
```
unindented: Latest stable is `6.11.3`. Should we keep up to date?

Dockerfile
swese44: if a tag isn't provided docker looks for "latest", i guess that's common for docker image naming. but the travis images don't have a latest so i had to specify the latest tag manually

.travis.yml
swese44: sure, updated

PR #40, opened 2017-09-07
Fix html diff

swese44:


swese44:
i think there might be a font issue with our monospace elements like pre/code and some inputs. I can't actually access a broken visual diff though, and i don't know why the travis would load different fonts if we're using the same container... closing, probably just not worth the effort

PR #43, opened 2017-09-11
SVG import task

swese44:
- Adds `yarn svg` script which converts all .svg files in /assets/Icons into React .jsx components in src/components/Icon/icons
- Adds a smoke test to ensure each generated icon component builds/renders
- Exports the existing Skype icon
- Upgrades some necessary dependencies + config
- Updates `.travis.yml` to only do branch commit runs on master
- [ ] Add license and eslint ignore header to each icon .jsx file

Notes:
- I had to add missing `viewBox` attributes to a few source .svg files; `viewBox` will be a required attribute when we get icons from Designers
- There are some peer dependency warnings but they aren't causing any problems


.travis.yml
```diff
@@ -9,4 +9,8 @@ cache:
   directories:
     - "node_modules"
 
+branches:
+  only:
+  - master
+
```
unindented: What's the effect of this?

assets/Icons/raw/lock.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="224" height="280"><path fill-rule="evenodd" d="M20 112h184c11.046 0 20 8.954 20 20v128c0 11.046-8.954 20-20 20H20c-11.046 0-20-8.954-20-20V132c0-11.046 8.954-20 20-20zM112 0c43.078 0 78 32.594 78 72.8V112H34V72.8C34 32.594 68.922 0 112 0zM66 112h92V78.4c0-25.626-20.595-46.4-46-46.4S66 52.774 66 78.4V112z"/></svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" width="224" height="280" viewBox="0 0 224 280"><path fill-rule="evenodd" d="M20 112h184c11.046 0 20 8.954 20 20v128c0 11.046-8.954 20-20 20H20c-11.046 0-20-8.954-20-20V132c0-11.046 8.954-20 20-20zM112 0c43.078 0 78 32.594 78 72.8V112H34V72.8C34 32.594 68.922 0 112 0zM66 112h92V78.4c0-25.626-20.595-46.4-46-46.4S66 52.774 66 78.4V112z"/></svg>
```
unindented: Is it normal that these three SVGs all have different dimensions?

package.json
```diff
@@ -102,6 +110,13 @@
     "storybook:static:stop": "pkill -f \"sh -c serve docs --port 5555\"",
     "visual:test": "backstop test --config config/backstop/config.js",
     "visual:approve": "backstop approve --config config/backstop/config.js",
+    "svg:clean-build": "rimraf build/icons",
+    "svg:clean-components": "del 'src/components/Icon/icons/*.jsx'",
```
unindented: `del`?

assets/Icons/raw/lock.svg
annr: No, we are waiting for updated SVGs that are square and specifically 240x240 I think.

src/components/Icon/icons/Attach.jsx
```diff
@@ -1,13 +1,9 @@
-/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
-/* tslint:disable:max-line-length */
-import * as React from 'react';
+import React from 'react';
```
annr: Wouldn't we need the copyright line in these files?

package.json
swese44: deletes individual files -- in this case the previously-generated .jsx react icon components, but not the index.ts file. is there a better way to do this?

.travis.yml
swese44: i had enabled travis to run when branches are updated, this tells travis to only run those for the master branch. so travis should run on PR updates as expected, and run against master after PRs are merged in. unless i misunderstood how this config works, the documentation wasn't super clear

src/components/Icon/icons/Attach.jsx
swese44: yes we do, there's a TODO npm script to add the file headers after the files are generated. i hadn't found an existing module to do that easily yet

PR #44, opened 2017-09-11
Allow excluding stories in visual diffs

annr:
This `hasStoryTypes()` tests first if storiesOf exists, and then if it does, test if the line containing storiesOf is commented-out (with any amount of whitespace between string test and forward slashes)


config/backstop/get-stories.js
```diff
@@ -10,7 +10,7 @@ const STORYNAME_REGEX = /\.add\s*\(\s*['|"].*?['|"],/g;
 
 
 function hasStoryTypes(text) {
-  return text.indexOf('storiesOf') > -1;
+  return text.indexOf('storiesOf') > -1 && text.match(/\/\/.\s?storiesOf/) === null;
```
swese44: it won't catch `/* ... */` comments, but then again this whole file is questionable :)

PR #45, opened 2017-09-11
add dropdown component

chrismohr:


src/components/Dropdown/README.md
```diff
@@ -0,0 +1,71 @@
...
+/>
+```
+---
+
+## Example implementation
+
+```
+import Dropdown from 'yamui/dist/components/Dropdown';
+
+class ControlledDropdown extends React.Component<any, any> {
```
unindented: If a component is not going to receive props, I'd use `{}` and not `any`.

Given that this component has state, I'd declare it, and not use `any` either.

(I know it's an example, but people copy-paste examples, and I don't want them to be spreading `any`s.)

src/components/Dropdown/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Properties
+
+| Prop | Type | Description |
+| ---- | ---- | ----------- |
+| options *| [IDropdownOption] | Items to be presented to the user. |
+| label | string | Text that will be rendered above the dropdown.
+| selectedKey | string | Use to specify an initially selected option. |
+| onChanged | function | Function to call when user changes the selected item. |
+| placeHolder | string | Input placeholder text. Displayed until option is selected. |
```
unindented: Why did we go with `placeHolder`? The built-in attribute that `<input>` has is `placeholder`.

src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { BaseComponentProps } from '../../util/BaseComponent/props';
+import './dropdown.css';
+
+export type DropdownOptionKey = string | number;
+
+export interface DropdownProps extends BaseComponentProps {
+  /**
+   * Items to be presented to the user.
+  **/
+  options: IDropdownOption[];
```
unindented: Our interface here is leaking an implementation detail (the fact that we use Fabric under the hood). I would at least alias this type.

Also, you probably need to export it too, or your consumers won't be able to type their options properly.

src/stories/Dropdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import withReadme from 'storybook-readme/with-readme';
+import Dropdown from '../components/Dropdown';
+import Block from '../components/Block';
+import { FixedGridRow, FixedGridColumn } from '../components/FixedGrid';
+
+const readme = require('../components/Dropdown/README');
+
+class ControlledDropdown extends React.Component<any, any> {
```
unindented: Again, don't use `any`, even for examples.

src/components/Dropdown/README.md
swese44: it's Fabric's prop name. there are a few places where we surface Fabric's names directly even when they are not perfect, like the `<Fabric>` component. would you suggest that in general we keep the API surface more similar, or write more abstraction code for better naming or APIs?

src/components/Dropdown/README.md
swese44: @chrismohr i can walk though this with you

src/components/Dropdown/index.tsx
swese44: we're already surfacing a few Fabric interfaces where it makes sense, not to say we can't change that though. and here our component is named Dropdown so the name doesn't really give anything away. i'd agree that `DropdownOption` would be a better interface name than `IDropdownOption`

src/components/Dropdown/README.md
unindented: Ah right, if it's Fabric that's chosen this name, and we're just forwarding the props, then it's not worth changing.

src/components/Dropdown/index.tsx
unindented: Yeah, maybe `export type DropdownOption = IDropdownOption` would be enough.

PR #51, opened 2017-09-14
Auto-generated SVG components (TSX)

annr:
#### Pull request checklist
- [x] Add svgo with the config file from P.J.'s branch
- [x] Setup simple yarn command to run all svg tasks: `yarn svg`

A couple things to do for the next branch:
- aria-hidden=true is gone. I'll return that in a dynamic way if necessary.
- return the create svg component tool.


unindented:
`create-svg-component` has no tests, and is all based on regexes, which makes me nervous.

package.json
```diff
@@ -26,6 +26,7 @@
     "chalk": "=1.1.3",
     "concurrently": "=3.4.0",
     "cpx": "=1.5.0",
+    "create-svg-component": "^1.1.0",
```
swese44: need an exact version with `=`

package.json
```diff
@@ -73,6 +74,7 @@
     "clean:build": "rimraf build/*",
     "copy:story-assets": "cpx \"src/stories/assets/**/*\" \"dist/stories/assets\"",
     "copy:readmes": "cpx \"src/components/**/README.md\" \"dist/components\"",
+    "create:svg:components": "svgcomp --dir=assets/Icons/optimized --output=src/components/Icon/icons",
```
swese44: can the command just be `svg`?

src/components/Icon/icons/ArrowUp.tsx
```diff
@@ -0,0 +1,16 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+/* tslint:disable:max-line-length */
+import * as React from 'react';
+
+const ArrowUp = function ArrowUp(props: any) {
+  return (
+    <svg width={240} height={240} viewBox="0 0 240 240" xmlns="http://www.w3.org/2000/svg" xmlnsXlink="http://www.w3.org/1999/xlink" {...props}>
+      <defs>
+        <path d="M59.92 84.29l64.076-64.077L102.783-1 13.107 88.677 2.5 99.283l100.283 100.283 21.213-21.213-64.146-64.147 139.887.067.013-29.917-139.83-.066z" id="a" />
```
swese44: this path has an id, looks like one other SVG also has it

src/components/Icon/icons/ArrowUp.tsx
```diff
@@ -0,0 +1,16 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+/* tslint:disable:max-line-length */
+import * as React from 'react';
+
+const ArrowUp = function ArrowUp(props: any) {
+  return (
+    <svg width={240} height={240} viewBox="0 0 240 240" xmlns="http://www.w3.org/2000/svg" xmlnsXlink="http://www.w3.org/1999/xlink" {...props}>
```
swese44: since these are inline svgs we should remove both xmlns and xmlns:xlink

src/components/Icon/icons/ArrowUp.tsx
```diff
@@ -0,0 +1,16 @@
...
+      <defs>
```
swese44: should remove the `<defs>` wrapper

src/components/Icon/icons/ArrowUp.tsx
```diff
@@ -0,0 +1,16 @@
...
+        <path d="M59.92 84.29l64.076-64.077L102.783-1 13.107 88.677 2.5 99.283l100.283 100.283 21.213-21.213-64.146-64.147 139.887.067.013-29.917-139.83-.066z" id="a" />
+      </defs>
+      <use fillRule="nonzero" transform="rotate(90 99.625 119.783)" xlinkHref="#a" />
```
swese44: should be using the path directly and remove this `<use>` element

assets/Icons/README.md
```diff
@@ -1,44 +1,17 @@
...
-Next, add the component file you created to src/components/Icon/icons/ and continue the steps for [Adding Icons](https://github.com/Microsoft/YamUI#adding-icons)
+3. Export the new icon from src/components/Icon/icons/index.ts
 
 4. Commit all these resources to a new branch.
 
 ### SVG requirements
  - square field of 240
- - includes viewBox, width and height
+ - viewBox required, height and width optional
  - no id or fill in path element. fill-rule ok.
```
swese44: if we use svgo to clean the SVGs we import you can delete line 16 as id and fill will be removed

assets/Icons/README.md
```diff
@@ -1,44 +1,17 @@
 # Source SVG icons used to build React components
 
-### Here we collect the raw (exported from Sketch) and optimized SVGs.
+## To add a new icon or icons
 
-1. Add raw SVGs to assets/Icons/raw folder
+1. Add prepared SVGs to assets/Icons/optimized folder
 
-2. Process SVGs with svgo
+2. Run `yarn create:svg:components`
```
swese44: can we use a simpler command like `yarn svg`?

src/components/Icon/icons/ArrowDown.tsx
```diff
@@ -0,0 +1,13 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
```
swese44: did you manually add these headers? if so we should also automate this in another sub task of `yarn svg`

swese44:
really nice to have these generated as typescript

swese44:
i bet we can remove the "UX Designer has approved visual changes" checklist item, it seems redundant since designers are required to give a LGTM on visual changes

package.json
annr: Sure! I thought I was following a pattern that you made :)

assets/Icons/README.md
annr: Sounds good to me.

src/components/Icon/icons/ArrowDown.tsx
swese44: (but i had not found a good solution yet either)

unindented:
Some concerns with https://github.com/annr/create-svg-component:
- It seems pretty tied to this use case (e.g. https://github.com/annr/create-svg-component/blob/304b62e918f24fdf91b7f275ebe085f12ee24b10/src/generateComponentFile.js#L10-L12).
- It has no tests.
- It's using the default `ISC` license.

Are you sure it's worth splitting out into it's own package? Why not take advantage of the fact that this repo is already approved?

docs/index.html
```diff
@@ -0,0 +1,51 @@
...
+          }
+        </style>
+        
+      </head>
+      <body style="margin: 0;">
+        <div id="root"></div>
+        <script src="static/manager.9c853e7e337a965e8214.bundle.js"></script>
+      </body>
+    </html>
+
```
unindented: All the stuff in the `docs` folder can now be removed.

.svgo.yml
```diff
@@ -0,0 +1,15 @@
...
+  - removeDimensions: true
+  - sortAttrs: true
+  - removeStyleElement: true
+  - removeScriptElement: true
+  - removeAttrs:
+      attrs: ['stroke', 'fill', 'style', 'xmlns', 'xmlns:xlink']
+
+js2svg:
+  pretty: true
+  indent: '  '
```
swese44: file should end with a newline

package.json
```diff
@@ -101,6 +103,9 @@
     "storybook:static:build": "build-storybook --config-dir config/storybook --output-dir build/storybook",
     "storybook:static:serve": "serve build/storybook --port 5555 || true",
     "storybook:static:stop": "pkill -f \"node_modules/.bin/serve build/storybook\"",
+    "svg:svgo": "svgo --folder=./assets/Icons/svgs-from-design --output=./build/icons --config=./.svgo.yml",
```
swese44: is there a reason we don't just have designers dump the SVGs directly into `assets/Icons` and pull from there?

swese44: since we're doing a full build of the assets/Icons directory each time i think we should also be cleaning the `build/icons` and `src/components/Icon/icons` directories to ensure we're not accidentally keeping any artifacts:

```yarn add -D del-cli```

```
"svg:clean-build": "rimraf build/icons",
"svg:clean-components": "del 'src/components/Icon/icons/*.tsx'",
```

Comment (unindented) [reply package.json]: We already have `rimraf`, no need for `del-cli`.

Comment (swese44) [reply package.json]: we need to remove all the component .tsx files but leave the index.ts file, i think rimraf just deletes directories. unless we want to generate the index.ts file as well, that would mean one less manual step of adding new icon imports

Comment (unindented) [reply package.json]: `rimraf` supports globs fine. No need for `del-cli`.

**Comment by mamuso** [comment]
If we are only interested in the clean SVGs, do we really need to keep the "svgs-from-design" folder?

File: package.json (hunk 12 above)
Comment (swese44) [reply package.json]: oh nice i was not aware

**Comment by annr** [comment]
@mamuso -- I wanted to be explicit about where the icons were coming from and also allow for other resources to be put into the assets/Icons folder. But both you and @swese44 don't see a reason for it, so I'll take the SVGs out of that folder. 

As for other comments in this PR....I think I'm fine with all of these suggestions and will update the PR when I get a chance.

File: src/components/Icon/icons/index.ts [hunk 13]
```diff
@@ -1,26 +1,45 @@
...
 export default {
+  arrowDown,
+  arrowLeft,
+  arrowRight,
+  arrowUp,
   attach,
+  carrotDown,
+  carrotLeft,
+  carrotRight,
+  carrotUp,
```
unindented: Why do we call these *carrot*? They surely are *carets*. 🥕🥕🥕🥕

annr: haha.

annr: Adrian is changing the Sketch file names, and I'm updating this branch now. Thanks for catching this, haha.

annr: (and he's going to give us a new minus svg but later)

package.json
```diff
@@ -101,6 +102,10 @@
     "storybook:static:build": "build-storybook --config-dir config/storybook --output-dir build/storybook",
     "storybook:static:serve": "serve build/storybook --port 5555 || true",
     "storybook:static:stop": "pkill -f \"node_modules/.bin/serve build/storybook\"",
+    "svg:clean-build": "rimraf build/icons",
+    "svg:clean-components": "rimraf src/components/Icon/icons/*.tsx",
```
unindented: I would move these two tasks up with the other `clean` tasks, and rename them `clean:icons` and `clean:components` maybe? That way all tasks that delete stuff are grouped together under the same prefix.

src/components/Icon/icons/ArrowDown.tsx
```diff
@@ -0,0 +1,15 @@
...
+/* tslint:disable:max-line-length */
+// This file was auto-generated -- do not edit. Use https://www.npmjs.com/package/create-svg-component
```
unindented: I'd probably delete this line from the generated files to avoid confusion.

src/components/Icon/icons/Down.tsx
```diff
@@ -1,14 +1,15 @@
...
 /* tslint:disable:max-line-length */
+// This file was auto-generated -- do not edit. Use https://www.npmjs.com/package/create-svg-component
+
 import * as React from 'react';
 
 const Down = function Down(props: any) {
   return (
-    <svg xmlns="http://www.w3.org/2000/svg" width="320" height="320" viewBox="0 0 320 320" aria-hidden="true" {...props}>
-      <path d="M160 195.774L64.072 99.846A19.9 19.9 0 0 0 49.96 94a19.9 19.9 0 0 0-14.114 5.846c-7.795 7.795-7.795 20.432 0 28.226L144.803 237.03c.33.384.68.76 1.043 1.124A19.895 19.895 0 0 0 160 244a19.897 19.897 0 0 0 14.154-5.846c.365-.365.712-.74 1.043-1.125L284.154 128.07c7.795-7.794 7.795-20.43 0-28.226A19.906 19.906 0 0 0 270.04 94a19.9 19.9 0 0 0-14.112 5.846L160 195.774z"/>
+    <svg viewBox="0 0 320 320" width={320} height={320} {...props}>
```
unindented: This file is still 320x320, is that expected?

src/components/Icon/icons/Info.tsx
```diff
@@ -1,14 +1,15 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 /* tslint:disable:max-line-length */
+// This file was auto-generated -- do not edit. Use https://www.npmjs.com/package/create-svg-component
+
 import * as React from 'react';
 
 const Info = function Info(props: any) {
   return (
-    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 320" height="320" width="320" aria-hidden="true" {...props}>
+    <svg viewBox="0 0 320 320" width={320} height={320} {...props}>
```
unindented: This file is also 320x320, expected?

src/components/Icon/icons/Leave.tsx
```diff
@@ -1,14 +1,15 @@
...
 /* tslint:disable:max-line-length */
+// This file was auto-generated -- do not edit. Use https://www.npmjs.com/package/create-svg-component
+
 import * as React from 'react';
 
 const Leave = function Leave(props: any) {
   return (
-    <svg xmlns="http://www.w3.org/2000/svg" width="320" height="320" viewBox="0 0 320 320" aria-hidden="true" {...props}>
-      <path d="M40.014 269.246L40 50c0-11.044 8.95-20 19.992-20H100.2c11.046 0 20 8.954 20 20s-8.954 20-20 20H79.987L80 250h20c11.046 0 20 8.954 20 20s-8.954 20-20 20H60.008c-.636 0-1.266-.03-1.887-.087C47.958 288.966 40 280.413 40 270c0-.252.005-.504.014-.754zM231.824 140l-35.955-35.868c-7.827-7.807-7.826-20.463 0-28.27 7.83-7.81 20.52-7.817 28.36-.02l56.088 55.793 17.016 17.097c6.253 6.28 6.217 16.428-.08 22.667l-17.037 16.873-56.01 55.875c-7.826 7.803-20.512 7.802-28.337-.002-7.826-7.807-7.827-20.462 0-28.27L231.83 180H126.22c-7.85 0-14.22-8.954-14.22-20s6.368-20 14.223-20h105.6z"/>
+    <svg viewBox="0 0 320 320" width={320} height={320} {...props}>
```
unindented: This file is also 320x320, expected?

PR #52, opened 2017-09-14
Use PureComponent everywhere

unindented:
> PureComponent is exactly the same as Component except that it handles
the shouldComponentUpdate method for you. When props or state changes,
PureComponent will do a shallow comparison on both props and state.
Component on the other hand won’t compare current props and state to
next out of the box. Thus, the component will re-render by default
whenever shouldComponentUpdate is called.

Read more at:
https://codeburst.io/when-to-use-component-or-purecomponent-a60cfad01a81


unindented:
@swese44 @chrismohr @annr please review.

PR #55, opened 2017-09-15
Change pkill pattern, and pin yarn v0.27.5

unindented:
The `sh -c serve docs` task is the parent of the `node_modules/.bin/serve docs` task. It's probably better to kill the child and let the parent terminate because of that. Also pkill may not find the process if it was started with a shell other than `sh`?

Also, we've seen that yarn v1.0.0 treats tasks that receive a SIGTERM as if they errored, which screws the way we run our visual tests. We're going to pin the version to v0.27.5 until we find a better solution.


unindented:
@annr @swese44 please review.

PR #56, opened 2017-09-15
Enable type-checking on tslint

unindented:
It'll also check for undeclared variables and other things.


unindented:
@swese44 @annr please review.

PR #58, opened 2017-09-18
Docker image bloat instructions

annr:
I don't like that we muck up our README instructions with this information, but I do think it is necessary. The wording I'm not opinionated about, and can be changed however, as long as we give people a warning. We can anticipate most (all?) YamUI devs having this issue eventually.


README.md
```diff
@@ -10,6 +10,12 @@ To run visual diff regression tests:
 * Install docker from https://docs.docker.com/docker-for-mac/install/
 * Create an alias `y` in your `~/.zshrc` or `~/.bashrc` file for the `yarn-docker` executable, like this: `alias y="./yarn-docker"`. Visual-diff tasks `test` and `test:visual` need to run in Docker via the `y` alias -- `y test` and `y test:visual`.
 
+A note about Docker images: they grow until your hard drive fills up, stopping you from working. Ideally you have a very large hard drive -- like 500GB -- or a lot of available space. If you don't have a lot of available space (> 30GB) you can expect to need to clear the Docker container every week or so.
```
unindented: Maybe we can phrase this less negatively? "If you find that Docker containers and images are taking too much space, try the following:"

annr: I like that better, but let's tell them how much free space is required. Here's the new text:

"It's best to have at least 30GB of free space for Docker containers and images. If you find that Docker is taking up too much space, try the following:"

PR #60, opened 2017-09-25
Add Spinner component

unindented:
A spinner is an outline of a circle which animates around itself indicating to the user that things are processing. It is shown when we're unsure how long a task will take.

#### Pull request checklist
- [x] UX Designer has approved visual changes


src/util/enums/spinner.ts
```diff
@@ -0,0 +1,10 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import { Enum } from 'typescript-string-enums';
```
unindented: Now that TypeScript supports string enums, we can remove this in a separate PR.

src/stories/assets/stories.css
```diff
@@ -0,0 +1,4 @@
+.animations-disabled * {
```
unindented: Suggestions for a better class name?

src/components/Spinner/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export interface SpinnerProps extends BaseComponentProps {
+  /**
+   * Label displayed below the spinner.
+   */
+  label?: string;
+  /**
+   * LARGE: 48px, MEDIUM: 40px, SMALL: 32px, XSMALL: 24px. Defaults to medium.
+   */
+  size?: SpinnerSize;
```
unindented: Waiting for Design to come back with all scenarios we need to support.

After that, I'll consult with @BrendanMcK, see how we want to deal with labels and ARIA.

src/components/Spinner/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```
+<Spinner size={ SpinnerSize.large } />
+```
+
+### Spinner with label
+
+```
+<Spinner label="Loading..." />
+```
```
unindented: @swese44 are these examples too obvious? What do we usually have as examples?

swese44: i think that's fine, there's not much logic going into a spinner

unindented:
Visual diffs are randomly failing one or both of the spinner tests. Seems like animations are not being disabled?

src/components/Spinner/README.md
```diff
@@ -0,0 +1,34 @@
...
+---
+
+### Properties
+
+| Prop | Type | Description |
+| ---- | ---- | ----------- |
+| label | string | Text that will be rendered below the spinner.
+| size | SpinnerSize | Size of spinner.
+
+* indicates property is required
```
annr: If there are no required properties we can remove this note.

unindented:
The increased thresholds for spinners seems to have solved the issue.

src/components/Spinner/README.md
unindented: Fixed.

unindented:
I'll still wait for Design to let us know all the different scenarios, cause I'll probably need to update things.

unindented:
@mamuso @annr I've updated according to spec. Please review.

unindented:
@adrianmg in case @mamuso is not available...

PR #61, opened 2017-09-26
Dropdown styles

bryan-gislason:
I've updated the styles as much as possible.

I still have to do the placeholder color, but I'm waiting for the latest fabric-ui update I've made to be available so that I can override that color.

screen shot 2017-09-25 at 5 54 12 pm
screen shot 2017-09-25 at 5 53 54 pm
screen shot 2017-09-25 at 5 53 50 pm
screen shot 2017-09-25 at 5 53 47 pm

#### Pull request checklist
- [ ] UX Designer has approved visual changes

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/yamui/61)


src/components/Dropdown/dropdown.css
```diff
@@ -2,8 +2,55 @@
...
+  line-height: 3.3rem;
+}
+
+.y-dropdown .ms-Dropdown-title {
+  color: $textColor;
+  border-color: $borderColor;
+}
+
+.y-dropdown .y-dropdown__caretDown {
+  color: $color__river;
```
bryan-gislason: Can anyone advise what we should do here? I'd like to have some of these colors mapped to names in our colors file, but they didn't exist there yet. Thoughts on all of the colors I directly used from the palette?

unindented: We seem to be using semantic color names everywhere else, so if there isn't a semantic color already, we should create one.

@mamuso @adrianmg is this a color we'll use just for carets (i.e. we name it `caretColor`) or will it be used for other types of things?

src/components/Dropdown/dropdown.css
```diff
@@ -2,8 +2,55 @@
...
+
+.y-dropdown .ms-Dropdown:hover .ms-Dropdown-title,
+.y-dropdown .ms-Dropdown:active .ms-Dropdown-title,
+.y-dropdown .ms-Dropdown:focus .ms-Dropdown-title {
+  color: $textColor;
+  border-color: $borderColor;
+}
+
+.y-dropdown .ms-Dropdown:active .ms-Dropdown-title {
+  border-color: $color__lake;
```
unindented: Why not name this too?

src/components/Dropdown/dropdown.css
```diff
@@ -2,8 +2,55 @@
...
+}
+
+.y-dropdown .ms-Dropdown:focus .ms-Dropdown-title {
+  border-color: $color__lake;
```
unindented: Same.

src/components/Avatar/avatar.css
```diff
@@ -24,7 +24,7 @@
   height: 24px;
   width: 24px;
 }
-.y-avatar__size-xSmall > .ms-Persona {
+.y-avatar__size-xSmall .ms-Persona {
```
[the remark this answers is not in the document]
annr: Basically Fabric does not allow avatars to be smaller than 32px, and this is our way to fix that. I think that the proper fix might be in Fabric. 

Also we can change the comment "extract a Fabric Avatar component out of Persona" since we never plan to do that.

src/components/Dropdown/dropdown.css
```diff
@@ -2,8 +2,55 @@
...
+}
+
+/* NOTE: y-dropdown cannot be used here since the options are not a descendent
+ * of dropdown element
+ */
```
annr: Reaching down into the Fabric styles feels bad, but that's what we've been doing and I know you're just following the pattern.

Looks like we can pass a className prop to Dropdown if that feels better to you.

src/components/Avatar/avatar.css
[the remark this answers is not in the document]
swese44: feel free to look at a fix within Fabric but it's not high priority, this style override is protected by visual diffs

extracting Avatar from Persona is still a viable project at some point -- it would reduce our amount of dependency JS from Fabric and other teams would benefit. also not high priority for us though

src/components/Dropdown/dropdown.css
```diff
@@ -2,8 +2,55 @@
...
-.y-dropdown .ms-Dropdown-label{
+.y-dropdown .ms-Dropdown-label {
   @extend text__mediumSub;
   font-weight: $fontWeight__bold;
   color: $textColor__secondary;
 }
+
+.y-dropdown .ms-Dropdown-title,
+.y-dropdown .ms-Dropdown-caretDownWrapper {
+  height: 36px;
```
swese44: question for @mamuso - should a Dropdown's height be set in `px` or `rem`? We use `rem` for fonts because they should scale (font + line-height/vertical rhythm) based on the user's default browser font-size. so should elements like Textfield and Dropdown also grow vertically if the text size has changed?

src/components/Dropdown/dropdown.css
swese44: you can just use $linkColor here, set it colors.css

src/components/Dropdown/dropdown.css
swese44: @mamuso looking at mockups from you and Kike, the focused Textfield's border color is our link color (river, #386cbb) while the focused Dropdown border color is lake (#6c98d9). Should these be consistent?

@bryan-gislason i'd recommend making a new variable in colors.css like `$focusedBorderColor` set to $color__lake. if @mamuso says focused input borders should be our link color then i'd still keep $focusedBorderColor as a separate variable and set its value to $linkColor

src/components/Dropdown/dropdown.css
swese44: maybe $focusedBorderColor ?

src/components/Dropdown/dropdown.css
```diff
@@ -2,8 +2,55 @@
...
+}
+
+.y-dropdown .ms-Dropdown:focus .ms-Dropdown-title {
+  border-color: $color__lake;
+}
+
+/* NOTE: y-dropdown cannot be used here since the options are not a descendent
+ * of dropdown element
+ */
+.ms-Dropdown-items .ms-Dropdown-item:hover,
```
swese44: @bryan-gislason maybe we need to also provide an `onRenderContainer` method, and put our `y-dropdown` class on the wrapper instead of passing it in as a className prop?

swese44: oh duh ignore this comment, we're already wrapping Dropdown in a div with our classname. so it looks like the items are rendered in a separate Layer? as @annr suggested maybe passing in a className to Dropdown would pass it along, or maybe there's another way to pass a className into the Layer wrapper?

swese44:
looking great! awesome to see the rendered component directly in the PR

src/components/Dropdown/dropdown.css
swese44: i think leave as-is, we can take a pass overall as Button might also need to change to `rem`

src/components/Dropdown/dropdown.css
swese44: and @mamuso has verified that the zeplin mocks are our source of truth, so in colors.css:

`$focusedBorderColor: $color__lake;`

bryan-gislason: should this be named something more like: borderColor--focused ?

src/components/Dropdown/dropdown.css
```diff
@@ -2,8 +2,55 @@
...
+.ms-Dropdown-items .ms-Dropdown-item:active,
+.ms-Dropdown-items .ms-Dropdown-item:focus {
+  background-color: $borderColor__tertiary;
```
bryan-gislason: Should we create some semantic names for$dropdownOption,  $dropdownOption--focused, $dropdownOption--selected ?

src/components/Dropdown/dropdown.css
swese44: ah good call

src/components/Dropdown/dropdown.css
swese44: for borders? i'd say no, probably better to leave them at a slightly higher abstraction so they can be shared by Textfield, etc.

bryan-gislason: should i just leave these as is using $borderColor__tertiary...?

swese44: that's what i would do. i supposed you could make the argument for `$inputBorderColor` variables, something like:

```
$borderColor: $color__indieRock;
$borderColor__secondary: $color__altRock;
$borderColor__tertiary: $color__popRock;

$inputBorderColor: $borderColor;
$inputBorderColor__focused: $color__lake;
```

src/css/variables/colors.css
```diff
@@ -10,6 +10,7 @@
 $borderColor: $color__indieRock;
 $borderColor__secondary: $color__altRock;
 $borderColor__tertiary: $color__popRock;
+$borderColor--focused: $color__lake;
```
swese44: please use underscores to keep consistent naming

src/components/Dropdown/dropdown.css
bryan-gislason: woops. we're talking border color, but it's background color here, but it's really specific to dropdown options. I think I should just use the colors directly here from the palet for the background colors for these dropdown options.

unindented: Maybe all those names are too specific. It'd be better to have a more generic name that we can use outside of dropdown.

bryan-gislason: Not that I disagree, but Fabric uses specific colors in their dropdown options from their palette.
```
&:hover {
    background-color: $Dropdown-selectedItem-hover-bg;
    color: $ms-color-black;

    @include highContrastListItemState;
  }

  @include focus-border();

  &:focus {
    background-color: $ms-color-neutralLighter;
  }

  &:active {
    background-color: $Dropdown-selectedItem-hover-bg;
    color: $ms-color-black;
  }

  &.itemIsDisabled {
    background: $ms-color-white;
    color: $ms-color-neutralTertiary;
    cursor: default;
  }
```

src/components/Dropdown/dropdown.css
bryan-gislason: the passed  in className unfortunately won't apply to the dropdown options since they are rendered in a separate layer. We could propose an additional className for options in Fabric?

src/components/Dropdown/dropdown.css
swese44: ok i looked at Drodpown and figured this out one. you can pass a `calloutProps` prop to Fabric's Dropdown like this:

```
<Dropdown
  ...
  calloutProps={{ className: 'y-dropdown--callout }}
```

then you could use the className you pass in your selectors

mamuso:
Stylewise we are a bit far from mocks:
- Vertical rhythm
- Font sizes
- Paddings

I'm pull-requesting a few of those changes soon and will come up with questions.

We are also missing the cases with icons, I guess that is something that we are not doing in this branch (or yet).

bryan-gislason:
@mamuso
I see that I need to update the options font-size from 15px and set to 1.4rem (14px).
I see some padding that's needs to be resolved (biggest one being the left-padding should be 12px, especially for the options).
I'm not sure about the vertical rhythm.

mamuso:
@bryan-gislason just pushed a few style changes, taking care of padding, rhythm, border radius and callout styles. What is missing before approval:

- Placeholder titles should be heavy metal
- For some reason, when we open the drowdown, the first option gets focused/selected and hovering over other options don't make it change

We should consider too.
Missign styles for:
- Disabled
- Error

Missing stories for:
- Options w/rich text
- Options w/icons

mamuso:
Also, the arrow down is not as design

bryan-gislason:
I used the Icon component with the 'down' icon. Is that icon intended for different purposes?

bryan-gislason:
@mamuso 
screen shot 2017-09-27 at 1 23 47 pm

bryan-gislason:
Also,

@mamuso

> Placeholder titles should be heavy metal
I'm working on this.

> For some reason, when we open the drowdown, the first option gets focused/selected and hovering over other options don't make it change
I'll look into this.

I'm holding off on the other missing styles/stories for now since they are not immediately needed for the new create group form.

bryan-gislason:
@mamuso, your 'overflow: hidden' in callout.css hid the 'beaks' and broke a bunch of visual tests. I've removed overflow:hidden. Perhaps this is something you want to add to on a descendent of the callout?

kukukaka:
@bryan-gislason I was checking the implementation for the dropdown in the group creation and I notice some differences. Sorry if they already have been solved.

- The carrot icon (I think this was already mentioned)
- The carrot icon color
- The separation between the watermark text and the input left border
- The border color
- The border radius
- The background color of the selected and hover states looks different from the proposal too.
- The shadow of the open list (I know this is usually tricky to align with the designs, but maybe we can tweak it?)

https://microsoft.sharepoint.com/:i:/t/yammerdesignsharing/EfsUpr3wLANCvaTETtgBzFEBcrhErxG-kP89-FezTi8wRg

Also, I was curious if we can have rich text in the dropdown options. I remember we talked about it, but I'm not sure about the current state.

Thanks!

src/components/Dropdown/index.tsx
```diff
@@ -51,6 +52,7 @@ export default class Dropdown extends React.PureComponent<DropdownProps, {}> {
           onChanged={this.handleChanged}
           placeHolder={placeHolder}
           label={label}
+          onRenderCaretDown={this.getIcon}
```
swese44: you can also add `calloutProps={{ className: 'y-dropdown--callout' }}`, so you can scope the list CSS rules to that parent

mamuso:
@bryan-gislason that was for keeping the border radius visible. We can use pseudo elements to make it work. Will take care of that, but in the meantime we can take the overflow out

mamuso:
@bryan-gislason  about the icon, I will update to the right one and will adjust. I think we have it right in one of the icon branches.

bryan-gislason:
@mamuso I've updated to include the new icon and also the placeholder text color. I can't seem to fix the issue where the option has an active background color when you click on the dropdown. I think it's an issue with fabric.

screen shot 2017-09-28 at 1 54 48 pm
screen shot 2017-09-28 at 1 54 43 pm
screen shot 2017-09-28 at 1 54 38 pm
screen shot 2017-09-28 at 1 54 31 pm
screen shot 2017-09-28 at 1 54 26 pm

mamuso:
Made a few adjustments:
- Fixing position of the carrot
- Fixing change of text color on hover when option is not selected

With that, we should be good to go. Let's not forget, please, the rest of the cases

PR #64, opened 2017-09-27
Add Microsoft OSS svgcomp which includes tests and doesn't use any regexes

annr:
I rewrote the package on my own time, kept it separate, and made it OSS under Microsoft. I didn't feel good about adding it to YamUI -- it's a separate tool and could be useful elsewhere, even Yamjs.

It's going through approval, but we don't have to merge this right away.

Tests cover the core functionality. If anyone thinks something should be added or changed, feel free to open a PR.

https://github.com/Microsoft/svgcomp


annr:
@unindented

PR #66, opened 2017-10-04
Lint React components using the `tslint-react` ruleset

unindented:
I ran `prettier` on all files, which solved most of the linting issues.

It aligns this repo with the linting rules we follow at Yammer.


unindented:
@annr @bryan-gislason please review.

PR #72, opened 2017-10-17
add icon label as class in the form 'yj-icon-label' in order to test …

annr:
Yamjs unit tests look for specific icons, but we don't currently pass that info.

### DIFF
--- a/src/components/Icon/__snapshots__/icon.test.tsx.snap
+++ b/src/components/Icon/__snapshots__/icon.test.tsx.snap
@@ -2,7 +2,7 @@
 
 exports[`<Icon /> with additional className matches its snapshot 1`] = `
 <Attach
-  className="y-icon TEST_CLASSNAME"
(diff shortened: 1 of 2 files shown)


annr:
@unindented -- what do you think of this solution for preserving Yamjs unit tests?

annr:
We probably want to do this in any case -- have a way to tell what icon it is from the markup, that is.

unindented:
Yup, I think having an attribute indicate this is a good idea. Maybe not the class name, in case we ever mangle it, with something like CSS modules (https://github.com/webpack-contrib/css-loader#modules) or Fabric's `merge-styles` (https://github.com/OfficeDev/office-ui-fabric-react/tree/master/packages/merge-styles)? What about a `data-` attribute?

annr:
What do you prefer? data-label or data-icon-label?

annr:
or data-icon, data-icon-name... please choose or offer another.

unindented:
`data-icon` sounds good to me.

PR #74, opened 2017-10-20
Messagebar Component

richardtagger:
Adds the message bar component - awaiting sign off from @adrianmg 

## Pull request checklist

* [x] Somebody from the UX team has reviewed your changes.


richardtagger:
@unindented @annr

unindented:
Why did all your reference files change? Only your feature should have changed right? Did you use docker?

richardtagger:
Not sure why that happened. Fixed

richardtagger:
Ok updated to Adrian's spec. Ready for review

src/components/MessageBar/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  if (!props.actions) {
+    return null;
+  }
+  return (<div className="y-message-bar--actions">{props.actions}</div>);
+};
+
+const MessageBar: React.StatelessComponent<MessageBarProps> = (props) => {
+  const actions = getActions(props);
+  return (
+    <div className={`y-message-bar y-message-bar__type-${props.type}`} aria-label={props.ariaLabel}>
```
unindented: All our components can take a `className`. You can do something like this: https://github.com/Microsoft/YamUI/blob/958d59a11d6035debf675379c434077c1d2c28f4/src/components/Callout/callout.tsx#L135

richardtagger: Alright - is that documented?

src/components/MessageBar/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  WARNING: 'warning',
+  ERROR: 'error',
+});
+export type MessageBarType = Enum<typeof MessageBarType>;
+
+export interface MessageBarProps extends NestableBaseComponentProps {
+  ariaLabel: string;
+  type?: MessageBarType;
+  actions?: JSX.Element;
+  children: JSX.Element | string;
```
unindented: I think these should not be `JSX.Element`, but `React.ReactNode[]`.

src/components/MessageBar/message-bar.css
```diff
@@ -0,0 +1,36 @@
...
+.y-message-bar__type-info {
+  border-left-color: $color__punkRock;
+}
+
+.y-message-bar__type-warning {
+  border-left-color: $color__noon;
+}
+
+.y-message-bar__type-error {
+  border-left-color: $color__angel;
```
unindented: You could alias these in the `colors.css` file, and have something like `$messageBarColor__error` in this file.

src/stories/MessageBar.tsx
```diff
@@ -0,0 +1,33 @@
...
+    </div>
+  ))
+  .add('with a type', () => (
+    <div>
+      <MessageBar ariaLabel="Error describing what went wrong" type={MessageBarType.WARNING}>
+        Watch out, something unexpected might happen.
+      </MessageBar>
+    </div>
+  ))
+  .add('with an action', () => {
```
unindented: Could we have another story with an icon action, like in Zeplin? Just to check that the spacing is all correct.

src/stories/gutters.css
```diff
@@ -0,0 +1,7 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+$gutterWidth__xxLarge: 2.4rem;
+$gutterWidth__xLarge: 2rem;
+$gutterWidth__large: 1.6rem;
+$gutterWidth__medium: 1.2rem;
+$gutterWidth__small: 0.8rem;
+$gutterWidth__xSmall: 0.4rem;
```
unindented: Why is this file here?

src/components/MessageBar/message-bar.css
richardtagger: mmm I like the locality of putting the message color in this css file. Do we really want styling info for this component in colors.css?

src/stories/gutters.css
richardtagger: no idea

src/components/MessageBar/index.tsx
unindented: Feel free to document it. I picked it up by looking at all the existing components, and the tests for the existing components. They all check that `className` can be passed.

src/components/MessageBar/message-bar.css
unindented: It's not the styling info. It's providing semantic names to colors. All component frameworks do this: Fabric, Bootstrap, Foundation, etc. Nobody litters their component CSS files with random color names, but with semantic names.

src/stories/gutters.css
unindented: Maybe it was an accidental drag-drop on your editor? Feel free to kill it.

src/components/MessageBar/message-bar.css
unindented: See for example https://github.com/twbs/bootstrap/blob/9a1d81f154f4391f9fea47fc26cd2fa864d6ae8a/scss/_buttons.scss#L81-L82

They create names for the colors, and then use the names.

src/components/MessageBar/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Examples
+
+### Basic usage
+
+```js
+const myIconComponent = <Icon icon="lock" />
+<MessageBar
+  ariaLabel="Warning that your post will be visible to some outside your network"
+  type={MessageBarType.Warning}
```
unindented: This'd be `MessageBarType.WARNING`.

src/components/MessageBar/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  INFO: 'info',
+  WARNING: 'warning',
+  ERROR: 'error',
+});
+export type MessageBarType = Enum<typeof MessageBarType>;
+
+export interface MessageBarProps extends NestableBaseComponentProps {
+  ariaLabel: string;
+  type?: MessageBarType;
+  actions?: JSX.Element;
```
unindented: This one should be `React.ReactNode` too.

src/components/MessageBar/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  children: React.ReactNode[] | string;
```
unindented: `React.ReactNode` includes `string`... Have you tried just doing `children: React.ReactNode`, without the array?

richardtagger: i just did what you said before.... i can change it to that

richardtagger:
ok - addressed all the things

PR #80, opened 2017-10-26
Turn all components into classes

unindented:
Function components don't have a display name, which was making things harder to debug for people.

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -23,39 +23,9 @@ export interface BaseButtonProps extends BaseComponentProps {
   ariaLabel?: string;
 
   /**
-   * Button size. Defaults to regular.
-   */
-  size?: ButtonSize;
-
-  /**
    * The color describing the button's intent. Defaults to primary.
    */
   color?: ButtonColor;
-
-  /**
-   * Click callback handler. Will provide the React synthetic event.
-   */
-  onClick?: (event: any) => void;
-
-  /**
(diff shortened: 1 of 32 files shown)


unindented:
@bryan-gislason please review.

src/components/Button/index.tsx
```diff
@@ -23,61 +23,61 @@ export interface BaseButtonProps extends BaseComponentProps {
...
+  onMouseLeave?: (event: any) => void;
 }
 
 export interface RegularButtonProps extends BaseButtonProps {
   /**
-   * Whether this button should be disabled or not. Defaults to false.
+   * Disabled buttons cannot have a link href.
    */
-  disabled?: boolean;
+  href?: void;
```
bryan-gislason: why is this void and not a string?

src/components/MediaObject/__snapshots__/media-object.test.tsx.snap
```diff
@@ -4,92 +4,92 @@ exports[`<MediaObject /> with additional className matches its snapshot 1`] = `
 <div
   className="y-media-object y-media-object__size-medium TEST_CLASSNAME"
 >
-  <Component
+  <FixedGridRow
```
bryan-gislason: Much better!

src/components/Button/index.tsx
unindented: We discussed in person. We think they're there to disambiguate the union.

PR #82, opened 2017-10-26
Add  textAlign to Block and update readme

swese44:
# Block alignment and README

- Adds textAlign: center for Block
- Adds missing textAlign props to README
- Gets test coverage back up to 100% by fully rendering a Dropdown component and verifying our caret icon override is displayed
- Removes a duplicate unit test in Dropdown
- Re-adds the `branches: master` filter in `travis.yml` to remove the extra run, @unindented confirmed that this was not the cause of the npm-auto-publish issue
- Removes the UX Designer checkbox from the PR template (already enforced by github codeowners)

## Pull request checklist


src/components/Block/block.test.tsx
```diff
@@ -61,21 +61,33 @@ describe('<Block />', () => {
...
     });
 
-    it('matches its snapshot', () => {
-      expect(component).toMatchSnapshot();
+    describe('center', () => {
+      beforeEach(() => {
+        component = shallow(<Block textAlign="center">block content</Block>);
+      });
+
+      it('includes the aligned right className', () => {
```
annr: Should be "includes the aligned center className."

swese44: oh duh, fixed

src/components/Dropdown/dropdown.test.tsx
```diff
@@ -79,6 +62,16 @@ describe('<Dropdown />', () => {
     });
   });
 
+  describe('when fully rendered', () => {
+    beforeEach(() => {
+      fullComponent = mount(<Dropdown options={[]} />);
+    });
+
+    it('displays our custom caret icon', () => {
+      expect(fullComponent.find('.y-dropdown__caretDown').length).toBe(1);
```
annr: It's not really testing for the icon, but a class. Maybe the desc could be more accurate.

swese44: this is the class we pass to the YamUI Icon component Dropdown renders, so by selecting this class we're selecting the Icon component's outer element. is there a different wording you'd prefer?

PR #84, opened 2017-10-26
Replace Storybook with Styleguidist

unindented:
- Replaces Storybook with Styleguidist
- Replaces BackstopJS with Puppeteer and a custom diffing script
- TravisCI runs now use our same Docker config, which should eliminate any transient hard-to-debug travis issues
- Removes the hacky server setup for visual diffs since the Styleguidist app runs fine from the filesystem

- [ ] OPIONAL, WE CAN IMPROVE THIS LATER. Should wrap the entire app in a `<Fabric>` component. Currently only stories are wrapped, so the page itself -- headings, descriptions, etc. -- are using Styleguidist's own styles, so our documentation doesn't entirely use the YamUI look and feel. We can't just wrap the whole app unfortunately, we'll have to override maybe the `StyleGuideRenderer` component? https://react-styleguidist.js.org/docs/cookbook.html#how-to-change-the-layout-of-a-style-guide Theming could be another option.

PR #86, opened 2017-10-29
Exploring broken visual diffs in Styleguidist branch

swese44:
Adds react-styleguidist dependency, the minimal change to break visual diff runs. Other new dependencies from that branch did not cause any problems.

https://github.com/Microsoft/YamUI/pull/84
@unindented

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -586,6 +586,29 @@
         }
       }
     },
+    "acorn-object-spread": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/acorn-object-spread/-/acorn-object-spread-1.0.0.tgz",
+      "integrity": "sha1-SOrQ9KjrFplaF6Dbn/xqyq2kumg=",
+      "dev": true,
+      "requires": {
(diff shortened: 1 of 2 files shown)


swese44:
@unindented it looks like react-styleguidist has a direct dependency on the `webpage` module so it's getting added to `node_modules`. casperjs has calls to `require('webpage').create()` -- `create` doesn't exist on that module, and casperjs doesn't even have a dependency on `webpage`. i'm guessing its `require` call to some local "webpage" module is getting hijacked by the npm module...

swese44:
looks like the `webpage` module is a phantomjs module
https://github.com/ariya/phantomjs/blob/master/src/modules/webpage.js

swese44:
removing the `webpage` module from our node_modules folder fixes the issue. it looks like react-styleguidist doesn't even use the webpage dependency, i'll try to verify that and see if they'll accept a pull request to remove it

swese44:
this should fix our issue
https://github.com/styleguidist/react-styleguidist/pull/655

PR #87, opened 2017-10-30
Updating MessageBar to reuse other components and reduce custom CSS

### LINKED ISSUE #3: Remove autoprefixer
BackstopJS uses a version of PhantomJS which doesn't support flexbox and some other CSS3 styles. We should keep an eye on this issue and see if we can remove autoprefixer when it's resolved:
https://github.com/garris/BackstopJS/issues/358

In the meantime we're using the same version and config as our client application

swese44:
# Updates to MessageBar component

- Reduces some MessageBar CSS by using `<Block>` for fonts and `<FixedGrid>` for layout
- Renames CSS classes like `$messageBarBorderColor__warning` to a more semantic/reusable `$alertColor__warning`
- Fixes an issue with FixedGrid components not taking up full width unless column contents pushed it full width
- Removes `ariaLabel` prop from MessageBar (breaking change)
- Removes unnecessary autoprefixer dependency

Fixes #3 

## Pull request checklist


src/components/MessageBar/README.md
```diff
@@ -8,7 +8,7 @@ Renders a component that displays relevant status information. You can use a Mes
 | Prop | Type | Description |
 | ---- | ---- | ----------- |
 | ariaLabel * | string | A description of the message bar for the benefit of screen readers. |
-| type | MessageBarType | Dictates the color scheme of the component. Defaults to `MessageBarType.INFO` |
```
richardtagger: why?

swese44: i think `status` better conveys the intent of this prop, `type` is a pretty overloaded term and requires me to look up the definition before i really know what it does.

happy to discuss though. this is just an experimental branch, i mainly wanted to see if i could reduce some CSS by reusing Block and FixedGrid inside MessageBar

richardtagger: I prefer type, if anything status implies state to me and they aren't stateful. I wouldnt make this change, especially cos its breaking, but if you want to go through with it and update YamJS I wont put up resistance

unindented: For what it's worth, Fabric's `MessageBar` also calls this concept `type`, although they use the redundant prop name `messageBarType`: https://developer.microsoft.com/en-us/fabric#/components/messagebar

unindented: I think I side with @richardtagger, `status` sounds like it's going to change at some point.

swese44: i can get behind that, i'll revert back to `type`.

actually i think it's the enum name `MessageBarType` that's bugging me. it's really not describing the type of message bar, it's describing the type/context of the message we're displaying -- which in the future could also be used in form errors/warnings, the flash message, maybe others. i'll leave it alone, but i think in the future when we add the FlashMessage component we'll want to decouple that enum from the MessageBar component and name it something like MessageType

unindented: Yup, sounds reasonable.

src/components/README.md
```diff
@@ -1 +1 @@
-&#8203;
+&#8203;
```
unindented: What's... the difference here?

swese44: i don't know... i accidentally deleted the file then copy+pasted the text back

src/components/Callout/Callout.md
```diff
@@ -143,37 +143,43 @@ const HintedPositionCallout = (props) => (
...
+        <HintedPositionCallout hint={DirectionalHint.leftCenter} align="left" label="Left center" />
+      </FixedGridColumn>
+      <FixedGridColumn fixed={true} width={100}>
+      </FixedGridColumn>
+      <FixedGridColumn fixed={true} width={100}>
+        <HintedPositionCallout hint={DirectionalHint.rightCenter} align="right" label="Right center" />
+      </FixedGridColumn>
+    </FixedGridRow>
+  </div>
+  <div>
```
unindented: Why are the extra `div`s necessary?

swese44: took a little head-scratching to figure out why the visual diffs were messed up for this story. basically i'd fixed a FixedGridRow bug via `flex-basis: 100%` because previously it would only fill its horizontal space if its column content pushed it out that far. in your example, your flex container's direct children were also display: flex and now have flex-basis: 100% so they filled the entire vertical space instead of being clustered in the center of the wrapper.

in general it's best to not use flexbox containers as direct children of another flexbox container, you should have another div or something to manage the flex-child layout then nest the next flex container under that

unindented: Aah I didn't stop to think that `FixedGridRow` was also `display: flex`.

unindented:
I think you just need a thumbs up from a designer.

PR #91, opened 2017-11-02
Change script back to /bin/bash

unindented:
`bin/sh` is not pointing to bash, so `set -o` fails.


unindented:
@swese44 please review.

PR #96, opened 2017-11-03
Provide retina screenshots

unindented:
The old screenshots were taken at 1x, and were blurry when viewed on a retina screen. @adrianmg didn't like that at all. These new screenshots are taken at 2x.


unindented:
@annr @swese44 please review.

unindented:
@adrianmg @mamuso please take a look when you have a minute. This just goes from 1x to 2x screenshots.

PR #98, opened 2017-11-03
Remove unused dependencies

unindented:


unindented:
@swese44 @annr please review.

unindented:
@swese44 We don't actually require the package anywhere.

unindented:
Ah, `@types/prop-types` is required by Fabric, cause they use it to set `context`. I'll add it back.

PR #101, opened 2017-11-06
[improvement] Update Clickable type to be button

richardtagger:
Currently using Clickable inside of forms will cause a submission as there is no type set, defaulting to ‘submit’ (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button)

# *Pull request subject*

*Describe the change you are making.*

## Pull request checklist


richardtagger:
@unindented @swese44

PR #102, opened 2017-11-07
Fix import paths in styleguide

unindented:
Import paths were not being generated properly after I moved the config file.

Here's what it was:

screen shot 2017-11-07 at 10 07 51

Here's what it should be:

screen shot 2017-11-07 at 10 07 56


unindented:
@swese44 @annr please review.

PR #103, opened 2017-11-07
Delay rendering of examples until they intersect the viewport

unindented:
When our styleguide page gets loaded, all callouts render even though they're not in view, and it looks janky.

screen shot 2017-11-07 at 11 49 55

I'm proposing we don't render an example until it enters the viewport, by using `IntersectionObserver`.


unindented:
@swese44 @annr please review.

PR #106, opened 2017-11-08
Auto-size Icon to parent Block/Text/Heading component if size is not set

### LINKED ISSUE #50: Consider adding codeowners to top-level codebase
We could add @annr, @unindented and @swese44 as code owners to all files. this would require one of us to give a LGTM, currently a Designer LGTM for the visual diffs would greenlight the entire PR

swese44:
Makes size optional and removes its default value, so an `<Icon>` can be used within any `<Text size={...}>` or `<Block textSize={...}>` and inherit the correct size and line-height. This behavior was requested from @bryan-gislason so I pulled this up from an old graveyard PR.

Also adds a missing example for Block's textSize.

Fixes #50 

## Pull request checklist


src/components/Icon/Icon.css
```diff
@@ -11,3 +17,36 @@
...
+  width: 2.1rem;
+  vertical-align: -0.1rem;
+}
+.y-textSize-large .y-icon {
+  height: 1.6rem;
+  width: 1.6rem;
+}
+/* medium matches the default values */
+.y-textSize-mediumSub .y-icon {
+  height: 1.4rem;
```
bryan-gislason: @swese44 would you consider using fontsize variables here instead so that these change relatively?

swese44: the icons change relative to the font-size from the nearest parent `<Text size={...}>` or `<Block textSize={...}>`. at a very local level this lets us use specific rem values to ensure crisp lines (instead of using em or %).

at a higher design level, we have a finite list of font-size + line-height combinations we support which match our vertical rhythm, we want people using YamUI components instead of writing their own CSS.

src/components/Button/Button.tsx
```diff
@@ -138,6 +139,7 @@ export default class Button extends React.PureComponent<ButtonProps, {}> {
     color: ButtonColor.SECONDARY,
     iconPosition: ButtonIconPosition.LEFT,
     size: ButtonSize.REGULAR,
+    status: ButtonStatus.ENABLED,
```
unindented: It's worth documenting that this is the default by adding a `@default` tag in the comment for the corresponding property.

src/components/Icon/Icon.css
```diff
@@ -1,13 +1,45 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 .y-icon {
-  display: inline-block;
-  /* IE/Edge do not bubble click events up from SVG elements, so allow clicks to pass through */
-  pointer-events: none;
+  display: inline;
```
unindented: Is this about vertical alignment?

swese44: yes

src/components/Button/Button.tsx
swese44: added

PR #107, opened 2017-11-09
 Make text mandatory on Spinner

unindented:
Spinners should always have text, for accessibility reasons. We do allow the text to be visible or not.


unindented:
@swese44 please review.

@adrianmg @mamuso we've renamed a field. The screenshots have changed in name but the content is exactly the same. We still need your thumbs up though.

src/components/Spinner/Spinner.md
```diff
@@ -64,30 +64,30 @@ const { SpinnerColor, SpinnerSize } = require('.');
...
-        <Spinner color={SpinnerColor.DARK} size={SpinnerSize.MEDIUM} label="Loading" />
+        <Spinner color={SpinnerColor.DARK} size={SpinnerSize.MEDIUM} text="Loading" showText={false} />
       </td>
       <td>
-        <Spinner color={SpinnerColor.DARK} size={SpinnerSize.SMALL} label="Loading" />
+        <Spinner color={SpinnerColor.DARK} size={SpinnerSize.SMALL} text="Loading" showText={false} />
       </td>
       <td>
-        <Spinner color={SpinnerColor.DARK} size={SpinnerSize.XSMALL} label="Loading" />
+        <Spinner color={SpinnerColor.DARK} size={SpinnerSize.XSMALL} text="Loading" showText={false} />
```
swese44: hrrmmm looking at this now i wonder if hideText is a better api, since text is required and would display by default? your call

unindented: I think you're right.

PR #108, opened 2017-11-10
Add loading status to Button

unindented:
Adds a loading status to `Button`.


unindented:
@swese44 @richardtagger please review.

@mamuso @adrianmg also please review.

src/components/Button/Button.css
```diff
@@ -65,12 +66,26 @@
...
 .y-button__state-disabled {
   opacity: 0.5;
   cursor: default;
 }
 
+.y-button__state-loading {
+  cursor: default;
+}
+
+.y-button__state-loading .y-button--contents {
```
swese44: whenever possible we want to keep selectors at 1 class-level selector, for DOM performance and specificity

src/components/Button/Button.css
```diff
@@ -65,12 +66,26 @@
...
+  visibility: hidden;
+}
+
+.y-button__state-loading .y-button--spinner {
```
swese44: this selector can just be `.y-button--spinner`

src/components/Button/Button.tsx
```diff
@@ -159,21 +183,26 @@ export default class Button extends React.PureComponent<ButtonProps, {}> {
...
-        className={this.getClasses()}
-        disabled={disabled}
-        href={href}
-        onBlur={onBlur}
-        onClick={onClick}
-        onFocus={onFocus}
-        onMouseEnter={onMouseEnter}
-        onMouseLeave={onMouseLeave}
-      >
+      <span className="y-button--contents">
```
swese44: you could add a class like `y-button--contents__hidden` if status==='loading' and target that with one CSS selector instead of two

unindented:
Sorry, I was too quick on the merge button.

src/components/Button/Button.css
unindented: Performance difference is insignificant, but agree wrt specificity.

src/components/Button/Button.css
unindented: True.

src/components/Button/Button.tsx
unindented: True.

swese44:
all good, super nitpicky

PR #109, opened 2017-11-10
Restrict push to [-3, 3] range

unindented:


src/components/MediaObject/MediaObjectMetadata.tsx
```diff
@@ -26,6 +26,7 @@ export default class MediaObjectMetadata extends React.PureComponent<MediaObject
       <Block
         className="y-media-object--metadata"
         textSize={TextSize.SMALL}
+        bottomSpacing={GutterSize.XSMALL}
```
swese44: i think this is breaking the visual diff as it's not conditionally adding bottomSpacing only if children were present. maybe we should just introduce topMargin on Block as well and leave the spacing attached to MediaObjectExtra?

unindented: Yeah, I tried to get away without conditionally adding the condition... Regarding `topMargin`, I'd rather not. It's nice to see most of the space in one direction.

unindented:
-

PR #111, opened 2017-11-10
Headings h1-h6

swese44:
Adds a Heading component. Renders as h1-h6, with size override option to render as h1-h6 or plain text.

## Pull request checklist


src/components/Heading/enums.ts
```diff
@@ -0,0 +1,3 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+export type HeadingLevel = '1' | '2' | '3' | '4' | '5' | '6';
+export type HeadingSize = HeadingLevel | 'none';
```
swese44: @unindented do you think these options make sense?

unindented: They made sense to me.

unindented: I guess why `none` instead of `null | undefined`?

swese44: default `undefined` value would use the current HeadingLevel style, so the unstyled text option needs an actual value. i think it's fine but it's a little odd

PR #112, opened 2017-11-14
Remove Components and Foundation folders, so Zeplin becomes source of truth

unindented:


unindented:
@swese44 @adrianmg please review.

PR #118, opened 2017-11-17
New Text and Block options

swese44:
- Adds `ellipsis` option to Block
- Adds `maxWidth` (with ellipsis), `bold` and `uppercase` options to Text
- Adds a new demo story to show `maxWidth` in our message reply use case
- [x] Update MessageReply demo links after https://github.com/Microsoft/YamUI/pull/106 is merged

@unindented 2 things:
1. any way to remove placeholder .tsx files for the .md demo examples?
2. i think webpack/uglify is removing whitespace characters from JSX which messes up spacing between tags like Text/span, would be nice to disable that


src/css/utility-classes.css
```diff
@@ -0,0 +1,13 @@
...
+  */
+
+.y-ellipsis {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  
+  /* overflow: hidden changes the baseline of inline-block elements, this fixes it */
+  vertical-align: bottom;
+}
```
unindented: Are `y-text*` classes also considered utility?

src/demos/MessageReply/MessageReply.md
```diff
@@ -0,0 +1,49 @@
...
+        <HorizontalListItem>
+          <Clickable><Icon icon="share" size={IconSize.XSMALL} /> Share</Clickable>
+        </HorizontalListItem>
+        <HorizontalListItem>
+          <Clickable>...</Clickable>
+        </HorizontalListItem>
+      </HorizontalList>
+    </Block>
+  </FixedGridColumn>
+</FixedGridRow>
```
unindented: Why not move all this into the `.tsx` file?

src/css/utility-classes.css
unindented: Ah sorry, that's happening in the other PR.

src/demos/MessageReply/MessageReply.md
swese44: good question. this way the full example is available for people to play with in the code panel.  would be nice if we didn't have to include a dummy tsx file

src/css/utility-classes.css
swese44: merged so now showing up here :)

src/demos/MessageReply/MessageReply.md
unindented: I understand. The code for `Hovercard` demos currently lives in the `.tsx` file though...

swese44: agree would be good to be consistent. let's think this through when we improve the examples/sections format

PR #124, opened 2017-11-22
Do not use ES6 features in action, as it doesn't get transpiled

unindented:
The arrow function was ending up in the bundle, causing IE to choke.


unindented:
@swese44 @bryan-gislason please review.

unindented:
@swese44 bump

PR #127, opened 2017-12-02
Upgrade office-ui-fabric-react to 5.29.0

chrismohr:
This is in preparation to implement TextField, since we'll need suffix from https://github.com/OfficeDev/office-ui-fabric-react/pull/3348

There were 3 changes to work around:
- Persona root size height is now calculated to accommodate a border indicating presence (https://github.com/OfficeDev/office-ui-fabric-react/pull/3291)
- Red is no longer in the PersonaCoin color palette (https://github.com/OfficeDev/office-ui-fabric-react/pull/3249)
- DirectionHint keys are no longer available in JS (https://github.com/OfficeDev/office-ui-fabric-react/pull/2984#pullrequestreview-66299384)


chrismohr:
I think we need to look closer at `config/styleguide-visual/reference/Callout_alignments_desktop.png`

It's slightly different now, however the original is not as I would have expected.

chrismohr:
@unindented, @swese44 please take a look.

swese44:
i think we can just remove the Callout Alignments example and its visual diff reference image, what do you think @unindented?

src/components/Avatar/Avatar.md
```diff
@@ -225,53 +225,53 @@ const { AvatarSize } = require('.');
...
-      <Avatar size={AvatarSize.MEDIUM} name="EEE EEE" initials="OR" />
+      <Avatar size={AvatarSize.MEDIUM} name="M" initials="PU" />
     </FixedGridColumn>
     <FixedGridColumn>
-      <Avatar size={AvatarSize.MEDIUM} name="FFF FFF" initials="DR" />
+      <Avatar size={AvatarSize.MEDIUM} name="N" initials="BL" />
     </FixedGridColumn>
     <FixedGridColumn>
-      <Avatar size={AvatarSize.MEDIUM} name="HHH HHH" initials="RE" />
+      <Avatar size={AvatarSize.MEDIUM} name="O" initials="TE" />
```
swese44: if i'm reading the Fabric PR correctly, they've removed RED and gone from 15 colors to 14. Can we update this example to remove the 15th avatar (duplicate color)?

src/components/Callout/Callout.md
```diff
@@ -3,7 +3,6 @@
 Default hover trigger:
 
 ```js { "props": { "data-example": "default hover trigger" } }
-const { DirectionalHint } = require('.');
```
swese44: is this not working anymore?

unindented: @chrismohr points it out in the description of the PR. They've gone from normal `enum`s to `const enum`s, which apparently don't generate JavaScript. They're citing reduced bundle size as the reason. So now we cannot use their constants from JavaScript!

unindented: As a solution, I was thinking of using the `updateExample` property, extracting all examples to their own separate files, and compiling them with `tsc`. That'd also solve the issue you cited of not being able to use your IDE to write examples. Whatchuthink?

swese44: @unindented are you suggesting that as part of this PR or as a follow-up item?

unindented: Let's say followup. I don't want this to be blocked.

chrismohr: https://github.com/Microsoft/YamUI/issues/131

PR #139, opened 2017-12-07
Design adding icons

mamuso:
We are adding and updating icons to our YamUI library

- [x] Fix TypeScript error at bottom of `Icon.test.tsx`
- [x] Add Icon documentation back into Styleguidist docs, `src/components/Icon/icons/Icon.md`
- [x] Make sure visual diffs have not changed since Manuel's last run


unindented:
@adrianmg I think you'll need to regenerate screenshots after having made that padding change.

src/components/MessageBar/MessageBar.css
```diff
@@ -10,6 +10,14 @@
   padding: 7px 12px 8px;
 }
 
+.y-message-bar__nowrap {
+  white-space: nowrap;
+}
+
+.y-message-bar__nowrap svg {
```
swese44: this seems dangerous, there could be an svg element in a Button with icon, or any other type of action component provided. does MessageBar itself just need an extra pixel of right padding?

src/components/MessageBar/MessageBar.tsx
```diff
@@ -39,7 +39,7 @@ export default class MessageBar extends React.PureComponent<MessageBarProps, {}>
       <Block textSize={TextSize.SMALL} className={this.getClasses()}>
         <FixedGridRow gutterSize={GutterSize.XXLARGE}>
           <FixedGridColumn>{children}</FixedGridColumn>
-          {actions && <FixedGridColumn fixed={true}>{actions}</FixedGridColumn>}
+          {actions && <FixedGridColumn fixed={true}><span className='y-message-bar__nowrap'>{actions}</span></FixedGridColumn>}
```
swese44: it might make more sense to handle this in FixedGridColumn. can you explain the problem this is solving?

src/components/MessageBar/MessageBar.css
```diff
@@ -7,7 +7,7 @@
   border-radius: 1px;
   border-left-width: 4px;
   color: $textColor;
-  padding: 7px 12px 8px;
+  padding: 7px 12px 7px;
```
swese44: this can now be:
`padding: 7px 12px;`

src/components/MessageBar/MessageBar.css
swese44: ah looks like the example has an icon next to text. if this is going to happen everywhere an icon is placed next to text then we should fix it at the Icon component level

src/components/MessageBar/MessageBar.tsx
mamuso: That line is introducing a couple of eslint warnings:
- ' should be "
- Exceeds maximum line length of 100

unindented: @swese44 should we pair on this today to fix any remaining issues, and get it out the door? That way people working in YamJS can start killing old icons.

adrianmg:
LGTM but it seems to need some fixes from @unindented and @swese44

swese44:
@unindented i've added some checklist items to the PR description

swese44:
@unindented i'm having trouble with the Icon examples. they display correctly in the full page, but styleguidist breaks when navigating to individual examples. any ideas?

unindented:
Does it log anything? I can try it tomorrow morning anyways.

swese44:
ya there is a javascript error in the console

PR #140, opened 2017-12-07
Introduce TextField Component

chrismohr:
## Pull request checklist


src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,134 @@
...
+        value={this.props.value}
+        className={this.getClasses()}
+        label={this.props.label}
+        description={this.props.description}
+        prefix={this.props.prefix}
+        suffix={this.props.suffix}
+        disabled={this.props.disabled}
+        required={this.props.required}
+        placeholder={this.props.placeHolder}
+        multiline={!!this.props.autoAdjustHeight || !!this.props.rows && (this.props.rows > 1)}
```
chrismohr: Right now multiline is implicitly set. This decision needs some more thought.

src/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,134 @@
...
+        rows={this.props.rows}
+        onGetErrorMessage={this.props.getErrorMessage}
+        deferredValidationTime={deferredValidationTime}
```
chrismohr: If we like doing validation with `onGetErrorMessage`, we should decide if `deferredValidationTime` should be a prop or if we just want to set a consistent value here.   The way I understand it, it's basically the debounce interval.

src/util/fabric/theme.ts
```diff
@@ -13,6 +13,12 @@ const fontStyles: IRawStyle = {
 
 // NOTE: Would be handy to have access to our CSS variable values in JS...
 loadTheme({
+  palette: {
+    themePrimary: '#6c98d9',
+    neutralSecondary: '#646d7a',
+    neutralTertiary: '#dde0e6',
+    neutralLighter: '#f3f5f8',
```
chrismohr: Setting colors here let me overcome some challenges with specificity of css selectors. However, I think we need a better way to managed these values.  When @unindented and I talked, we thought it might be ideal to declare our named colors in a json file, and allow that to be included here. This json file would then also replace colors.css. Maybe as part of the build we could generate colors.css from colors.json.

src/components/TextField/TextField.md
```diff
@@ -95,17 +95,17 @@ Required:
...
-  label="US Airport Code"
-  getErrorMessage={(value) => {
-	  return new Promise((resolve) => {
-      setTimeout(() => {
-        if (value.length <= 3) { resolve(''); }
-        resolve(`The length of the input value should less than 3, actual is ${value.length}.`);
-      }, 200);
-    });
-  }}
+  onChange={() => {console.log('change from onChange example')}}
```
swese44: i think you can use `action` here instead

src/components/TextField/TextField.tsx
```diff
@@ -100,19 +98,26 @@ export default class TextField extends React.PureComponent<TextFieldProps, {}> {
...
         resizable={false}
         autoAdjustHeight={this.props.autoAdjustHeight}
         underlined={this.props.underlined}
-        onChanged={this.props.onChanged}
+        onChanged={this.getOnChange()}
       />
     );
   }
 
+  private getOnChange() {
```
swese44: we're essentially executing this method every time props change (which would generally be after each keystroke in a controlled component). i think i'd prefer to pass a function directly to onChanged instead of executing this get function on every render cycle.

chrismohr: Thanks @swese44 , I was wondering about that too, and could probably benefit from some guidance here. Here are some alternatives I see:

1. Simply move this logic to the constructor. Downside is that if for some reason the consumer wants to change the `onChange` or `onChangeDebounceTime` props, it won't be possible. 

2. In addition to #1, apply this logic in componentWillUpdate, behind a check to see if either `onChange` or `onChangeDebounceTime` have changed.

chrismohr: Actually... disregard this, I think I was overcomplicating things, and understand what you mean now.

chrismohr: Scratch that, I'm back to where I was before. 
Do you think it's worth the complexity to do something like this?  Which part of the function looks the most costly to you?  Might some sort of perf test help us decide?

```js
  private onChange: (newValue: string) => void;

  constructor(props: TextFieldProps) {
    super();
    this.async = new Async(this);
    if (props.onChangeDebounceTime && props.onChange) {
      this.onChange = this.async.debounce(props.onChange, props.onChangeDebounceTime);
    }
  }

  componentWillUpdate(nextProps: TextFieldProps) {
    if (nextProps.onChangeDebounceTime && nextProps.onChange
      && (nextProps.onChangeDebounceTime !== this.props.onChangeDebounceTime
        || nextProps.onChange !== this.props.onChange)) {
      this.onChange = this.async.debounce(nextProps.onChange, nextProps.onChangeDebounceTime);
    }
  }

  render() {
    return (
      <FabricTextField
        value={this.props.value}
        className={this.getClasses()}
        label={this.props.label}
        description={this.props.description}
        prefix={this.props.prefix}
        suffix={this.props.suffix}
        disabled={this.props.disabled}
        required={this.props.required}
        errorMessage={this.props.errorMessage}
        placeholder={this.props.placeHolder}
        multiline={!!this.props.autoAdjustHeight || !!this.props.rows && (this.props.rows > 1)}
        rows={this.props.rows}
        resizable={false}
        autoAdjustHeight={this.props.autoAdjustHeight}
        underlined={this.props.underlined}
        onChanged={this.onChange}
      />
    );
  }
```

src/util/fabric/theme.ts
swese44: 👍

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,69 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+@import "src/css/variables/colors.css";
+@import "src/css/variables/fonts.css";
+
+.y-text-field .ms-Label {
```
swese44: should probably add a comment saying this is temporary tech debt until we can set styling/theme options for all Fabric label instances (that work is in progress)

src/components/TextField/TextField.md
```diff
@@ -0,0 +1,111 @@
+### Examples
+
+One Line:
+```js
```
swese44: each of these lines needs a data attribute for correct diff image naming, check the other .md files for examples

src/components/TextField/TextField.tsx
swese44: this looks like the way Fabric's textfield does it, though they don't worry about updating a new update value so this is slightly improved. does this work as expected when providing a value (controlled component) and a debounce value?

swese44: would probably also need to clear the async timeout on `componentWillUnmount`

swese44: which looks to be `async.dispose()`:
https://github.com/OfficeDev/office-ui-fabric-react/blob/4b1468da53a1c3f5e0ddf5750b31c8500237ce6b/packages/utilities/src/Async.ts#L34

swese44:
looks like there are a few numbered textfield reference images which can be deleted now that you've added example data-attributes. i think you have to delete the reference image, its .new image, and its .diff image, otherwise @unindented's styleguide-visual tool may recreate the source image

swese44:
i'd also add an example next to a Button to ensure the heights correctly line up

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,65 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+@import "src/css/variables/colors.css";
+@import "src/css/variables/fonts.css";
+
+// work is in progress to set styling/theme options for all Fabric label instances
+.y-text-field .ms-Label {
+  @extend text__mediumSub;
+  font-weight: $fontWeight__bold;
+  color: $textColor__secondary;
+  padding-bottom: 4px;
```
swese44: any paddings/margins which account for our vertical rhythm should be specified in rems, this way if a user changes their browser's default font size we can adjust all of the text and the "invisible text grid" automatically

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.y-text-field .ms-TextField-fieldGroup {
+  line-height: 3.3rem;
+  height: 36px;
```
swese44: 3.6rem

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.y-text-field.ms-TextField--multiline .ms-TextField-field {
```
swese44: can we remove the multiline code from Textfield now that the components are being separated?

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.y-text-field.ms-TextField--underlined .ms-TextField-field {
+  padding-left: 0px;
+}
+
+.y-text-field .ms-TextField-field {
+  @extend text__mediumSub;
+  color: $textColor;
+  padding: 0 8px;
```
swese44: 0 .8rem

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.y-text-field--with-prefix.ms-TextField--underlined .ms-TextField-field,
+.y-text-field--with-prefix .ms-TextField-field {
+  padding-left: 4px;
```
swese44: .4rem

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.y-text-field--with-suffix.ms-TextField--underlined .ms-TextField-field,
+.y-text-field--with-suffix .ms-TextField-field {
+  padding-right: 4px;
```
swese44: .4rem

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.y-text-field .ms-TextField-prefix,
+.y-text-field .ms-TextField-suffix {
+  padding: 0 8px;
```
swese44: 0 .8rem

src/components/TextField/TextField.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.y-text-field .ms-TextField-errorMessage {
+  @extend text__small;
+  color: $textColor__error;
+}
+
+/* Hide the asterisk that comes with required=true */
+.y-text-field .ms-TextField-fieldGroup::after {
+  display: none;
```
swese44: looks like this isn't getting properly hidden in the visual diff

chrismohr: In the case that there is no label, fabric shows an asterisk after the field. This hides that. 

The asterisk you saw is beside the label, I'll make the change to hide that too.

src/components/TextField/TextField.css
chrismohr: yup, removed.

unindented:
@chrismohr why doesn't this example fill its parent container?

textfield_beside_button_desktop

chrismohr:
@unindented I've now updated the example with the button to fill the full width

unindented:
@chrismohr Sorry, I didn't see this yesterday...

chrismohr:
With @mamuso's help, I just noticed that I had missed coloring the borders properly. (https://github.com/Microsoft/YamUI/pull/172#pullrequestreview-84327362)

The error border color should now be correct. (`#d1423b`)

Also, fabric by default was darkening the border color on hover, which I've now suppressed. Please let me know if we do actually want a different border color on hover. 

cc/ @adrianmg

mamuso:
@chrismohr, is there any way to color the placeholder text with $textColor__metadata?

chrismohr:
@mamuso thanks for reviewing.  I'll try to address these two points after lunch.

chrismohr:
@mamuso The placeholder color looks to be `#646d7a` which is `$textColor__metadata: $color__heavyMetal;`

What am I missing?
yamui_v3_0_2

chrismohr:
Using the digital color meter app on my mac, with the reference screenshots in the PR gives me `#656E7A`, which seems pretty close.

chrismohr:
@mamuso The underlined variation will now get the focus color.

mamuso:
@chrismohr in my branch, the placeholder color is #A9A9A9

image


I can see the blue line on focus for the other case

chrismohr:
@mamuso weird... I wonder if it's a browser/os issue since the reference images generated in the docker environment seem to be showing the correct color.  I'll dig in further. 
What browser are you using?

chrismohr:
I'm also seeing that the reference images haven't been updated since ac76b73, I'll do that now.  Maybe that will help to explain things.

chrismohr:
Looks like the reference image changes were just some the alignment/spacing changes, not color.  I'll keep trying to understand why @mamuso and I are seeing different placeholder colors.

swese44:
@mamuso we cannot reproduce what you are seeing, the placeholder text is always $textColor__metadata. visual diffs are also correct. our global reset CSS sets this for us, and we don't see any CSS in the page which would conflict. do you have any browser extensions which might be adding styles to the page?

can we approve this and push it out to unblock @chrismohr's project work, and track this in a github issue?

mamuso:
Found the issue, for some reason, my chrome in this computer is stuck in v56 (????), therefore it needs browser specific pseudo elements to work :/ 
image

PR #142, opened 2017-12-08
Move from PureComponent to Component

unindented:
PureComponents don't play well with `react-router` and its use of
`context`:
https://github.com/ReactTraining/react-router/blob/d1071993e6262bc2e752c79b52a445979ce67111/packages/react-router/docs/guides/blocked-updates.md

Fabric also uses Component for everything.

Proposal is to use Component by default, and only use PureComponent for specific things that need the perf optimization.

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -55,7 +55,7 @@ export interface AvatarProps extends BaseComponentProps {
 /**
  * An `Avatar` shows a thumbnail representation of both an individual or group.
  */
-export default class Avatar extends React.PureComponent<AvatarProps, {}> {
+export default class Avatar extends React.Component<AvatarProps, {}> {
   static defaultProps: Partial<AvatarProps> = {
     borderType: AvatarBorderType.ROUND,
     size: AvatarSize.MEDIUM,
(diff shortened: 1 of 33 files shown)


unindented:
@swese44 please review.

@rese @chrismohr @bryan-gislason @mrfunkycold @richardtagger you won't be affected, but you may be surprised when you see this in YamUI, so heads up.

src/components/Avatar/Avatar.tsx
```diff
@@ -70,7 +70,7 @@ export type AvatarProps = ImageAvatarProps | InitialsAvatarProps;
 /**
  * An `Avatar` shows a thumbnail representation of both an individual or group.
  */
-export default class Avatar extends React.PureComponent<AvatarProps, {}> {
+export default class Avatar extends React.Component<AvatarProps, {}> {
```
swese44: can we leave leaf node components like Avatar to use PureComponent?

unindented: I think it's simpler advice to give to people working in this repo to just use `Component`. Our `shouldComponentUpdate` boundary will probably be at the components connected to the store. We can measure the difference, but I think PureComponent won't give us anything at these leaf nodes.

swese44: sorry if you're waiting for my reply, still a LGTM from my end

unindented: Ah right, I forgot about this one.

PR #143, opened 2017-12-09
Update Avatar to remove initials

richardtagger:
Pass through to Persona’s intials implementation

## Pull request checklist


richardtagger:
@unindented @swese44

swese44:
i've updated the avatar colors example with names that resolve to the same colors as before. there was also some other visual diff noise, not sure why those reference images changed but @chrismohr had also noticed this on his branch so i figured we'd check these in to resolve it

richardtagger:
Thanks all

PR #144, opened 2017-12-09
[SuggestionsList] presentational component work

richardtagger:
Introducing the presentational components that will be used in suggestions list

## Pull request checklist


src/components/SuggestionsList/SuggestionsListItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+                size={MediaObjectSize.SMALL}
+                imageContent={avatar}
+                titleContent={title}
+                metadataContent={description}
+              />
+            </FixedGridColumn>
+            {horizontalSpacer}
+          </FixedGridRow>
+        </Block>
+      </Clickable>);
```
richardtagger: @swese44 this feels like a ton of components for something quite simple, no?

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -0,0 +1,158 @@
...
+      <span>
+        <Block bottomSpacing={GutterSize.XSMALL} />
+        <Block
+          padding={GutterSize.MEDIUM}
+          textAlign="center"
+          className="y-suggestions-list--status"
+        >
+          {content}
+        </Block>
+        <Block bottomSpacing={GutterSize.XSMALL} />
```
richardtagger: @swese44 I can't get this to be the right size - according to @adrianmg's design this should be 48px tall

src/components/SuggestionsList/SuggestionsListItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export default class SuggestionsListItem extends React.PureComponent<SuggestionsListItemProps, {}> {
+  render() {
+    const { isSelected, isHovered, name, search, imageUrl, description } = this.props;
+    const avatar = (
+      <Avatar
+        imageUrl={imageUrl}
+        name={name}
+        size={AvatarSize.SMALL}
+        initials="" // !!
```
richardtagger: This is being removed https://github.com/Microsoft/YamUI/pull/143

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -0,0 +1,158 @@
...
+   */
+  noResultsText: string;
+  /**
+   * The groups of items to render.
+   */
+  groupedItems?: SuggestionItemGroupProps[];
+  /**
+   * The index of current keyboard selection.
+   */
+  selectedIndex: number;
```
swese44: selectedId would be easier for the client to deal with when there are multiple groups of results. but it'd still be nice if the client didn't have to deal with the selection at all. would it be possible for SuggestionsList to accept a target prop which is a ref to the input or textarea, and listen to its keyboard interaction directly?

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * The active search that produced the current state.
+   */
+  search: string;
```
swese44: would `searchText` be more clear?

richardtagger: sure

src/components/SuggestionsList/SuggestionsList.tsx
richardtagger: So in a Draft world there's various props - onUpArrow, onDownArrow, handleReturn etc - how would it work?

src/components/SuggestionsList/SuggestionsListItem.tsx
swese44: i'd kill the fixed grid and add an extra 4px horizontal margin on the MediaObject. i will separately ask Design if these can have consistent padding instead

src/components/SuggestionsList/SuggestionsList.tsx
swese44: is there a way to get the ref to Draft's underlying component? haven't verified but might be possible?
https://github.com/facebook/draft-js/issues/1398#issuecomment-331047429
https://jsfiddle.net/js1vn56m/

swese44: if so, could pass that ref as SuggestionList's input target and listen to the events directly

src/components/SuggestionsList/SuggestionsList.tsx
swese44: ya the empty Blocks are ugly. i'd just add CSS where it's easier than wonky markup and we can see if we can build those options into Block etc. in the future

src/components/SuggestionsList/SuggestionsList.tsx
richardtagger: you don't want to be in the business of intercepting or interfacing with the DOM directly - that's the whole point in Draft. Moreover, Draft's event handling for return is handleReturn rather than onReturn - because it expects you to return a 'handled' vs 'not-handled' string

unindented: Maybe we can create a DOM element on mount, move it on render, and attach the callout to that element?

richardtagger: sketching some ideas up and will report back

richardtagger: https://github.int.yammer.com/gist/rtagger/645580bfd6de1f36bc11e31ddc929139 thoughts on this?

swese44: Draft says the up/down keys are non-cancelable:
https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditorProps.js#L156
If listening to those events on a Draft instance to handle up/down in SuggestionsList, does that mean the cursor position will be bouncing around in the textarea as well? if so, that's an argument for passing the DOM node ref to SuggestionsList to listen and cancel events while SuggestionsList is rendered.


Regarding selectedId vs. selectedIndex. Draft won't know or care about SuggestionsList or its selected Suggestion, your component is only using Draft APIs to place the SuggestionsList over the cursor and then inject the selected item into the editor. your component would be managing the data powering the list of suggestions and selected state. i'm suggesting to select by id instead of index because your data model powering the SuggestionsList could change at any point and it would be easier to make sure the selected item remains selected if you select it by id, especially once we implement grouped suggestions.

but i don't think we need selectedIndex/selectedItem at all if this isn't a controlled component. if we can pass SuggestionsList the input DOM element then it can listen to up/down keys itself and the parent component can just listen to the onItemSelected() callback prop

swese44: even though it says non-cancelable it passes the event object, so maybe you still can preventDefault()

richardtagger: Yeah there was no jumping around, I prevent defaulted here https://github.int.yammer.com/yammer/yamjs/pull/9494/files#diff-551493f5e3d5008270b2ae82f390f960R97

richardtagger:
@swese44 @unindented Ready for review - the only unsolved piece is what to do in Styleguidist - Callout is creating a div on document and escaping the box I'm trying to demo the component in, as soon as you scroll everything looks messed up

src/components/SuggestionsList/SuggestionsList.css
```diff
@@ -0,0 +1,23 @@
...
+.y-suggestions-list {
+  width: 280px;
+}
+
+.y-suggestions-list--with-status .y-suggestions-list--loading {
+  height: 52px;
+}
+
+.y-suggestions-list--with-status.y-suggestions-list--with-results .y-suggestions-list--loading {
+  height: 48px;
```
unindented: Should all these pixel measurements be in rems, so that they adapt for people with non-default font size?

unindented:
@richardtagger how did you fix the font mismatch issues you were having?

src/components/SuggestionsList/SuggestionsList.css
```diff
@@ -0,0 +1,23 @@
...
+.y-suggestions-list {
+  width: 280px;
+}
+
+.y-suggestions-list--with-status .y-suggestions-list--loading {
+  height: 52px;
+}
+
+.y-suggestions-list--with-status.y-suggestions-list--with-results .y-suggestions-list--loading {
+  height: 47px;
```
mamuso: On the design, this is 48px including the line on the top of the row, hence 47.

richardtagger:
@unindented I'm not 100% as I was trying loads of stuff but I think running `fc-cache -f -v` fixed it

src/components/SuggestionsList/SuggestionsList.css
richardtagger: ok - ill regenerate the reference files

richardtagger:
@mamuso yeah I dont really know how to solve this - Callout does nasty stuff to the DOM and I cant think of way around this as it takes the component out of the hierarchy and places it in a new node on BODY that it controls :(. If you click through to individual example view they look fine, but in the All view they look screwy and break massively on scroll. Waiting to see if @swese44 or @unindented might have a good idea about how we can fix that

src/components/SuggestionsList/SuggestionsList.css
```diff
@@ -0,0 +1,23 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+@import "src/css/variables/colors.css";
+
+.y-suggestions-list {
+  width: 280px;
```
unindented: Should we be using rems for all these dimensions?

src/components/SuggestionsList/SuggestionsList.md
```diff
@@ -0,0 +1,120 @@
...
+</div>
+```
+
+With results, when loading:
+
+```js { "props": { "data-example": "When Results" } }
+<div style={{ minHeight: '297px' }}>
+  <span id="suggested-list-4" />
+  <SuggestionsList
+    target="#suggested-list-4"
```
unindented: Using string refs will get removed at any point: https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

Please don't use them even in examples.

src/components/SuggestionsList/SuggestionsList.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+      target,
+      isLoading,
+      groupedItems,
+      onItemSelected,
+      loadingText: 'loadingText',
+      noResultsText: 'noResultsText',
+      selectedId: 2,
+      searchText: 'searchText',
+    };
+  };
```
unindented: This feels dirty. An impure function that relies on the outer values to be modified by the `beforeEach` blocks?

Better:

```js
// whatever values you think make sense here
const defaultProps = {
      target: null,
      isLoading: false,
      groupedItems: [],
      onItemSelected: () => {},
      loadingText: 'loadingText',
      noResultsText: 'noResultsText',
      searchText: 'searchText',
      selectedId: null,
};

const getProps = (overrides) => {
    return {
      ...defaultProps,
      ...overrides,
    };
};
```

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -0,0 +1,155 @@
...
+   */
+  searchText: string;
+  /**
+   * Called when an item in a group is selected by clicking.
+   */
+  onItemSelected(item: SuggestionItem): void;
+  /**
+   * The target that the Callout should try to position itself based on.
+   */
+  target: HTMLElement;
```
unindented: I'd sort these attributes a bit better. Mandatory attributes at the top, then optional attributes, then even handlers (`onItemSelected`).

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -0,0 +1,155 @@
...
+const withResultsClass = 'y-suggestions-list--with-results';
+const withStatusClass = 'y-suggestions-list--with-status';
+
+/**
+ * A `SuggestionsList` displays a list of search results in a dropdown.
+ */
+export default class SuggestionsList
+  extends React.PureComponent<SuggestionsListProps, SuggestionsListState> {
+
+  public state: SuggestionsListState = { hoveredId: null };
```
unindented: Why would the state be public?

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private getGroupResults = (group: SuggestionItemGroupProps) => {
+    const items = group.items.map(this.getResultItem);
+    return (
+      <li key={group.title}>
+        <ul onMouseOut={this.onMouseOut}>
+          {items}
+        </ul>
+      </li>);
```
unindented: Can you generate an example screenshot with multiple groups of results with their corresponding titles? I only see a single group with no title in the screenshot.

richardtagger: titles has been punted on - we wont be supporting it until we do the grouped search results and who knows when that'll ever be

src/components/SuggestionsList/SuggestionsList.tsx
richardtagger: screen shot 2017-12-14 at 9 29 34 am

richardtagger: looks like PureComponent has it public

src/components/SuggestionsList/SuggestionsList.test.tsx
richardtagger: lol oops i always do this

src/components/SuggestionsList/SuggestionsList.tsx
unindented: Umm, I don't like seeing a `public` there, and also having to repeat the type of state.

Anything against constructor?

richardtagger: nah dont care

src/components/SuggestionsList/SuggestionsList.md
richardtagger: mm this isnt a ref like the react property ref - this is target which goes to Callout target

unindented: You sure it's not using a React `ref` under the hood?

richardtagger: idk - what would you change this example to? Also - any ideas how to fix the stupid screwy Callout thing, I dont know a good solution.

unindented: I'd check Fabric's examples, but maybe it'll look like this?

```
<span ref={(target) => { this.target = target; }} />
<SuggestionsList target={this.target} />
```

unindented: I'll checkout your branch, see what's the screwy behavior.

unindented: `doNotLayer` causes the Callout to scroll with the page, as it doesn't absolutely position the thing:

```js
      <Callout
        target={this.props.target}
        className={classNames.join(' ')}
        isBeakVisible={false}
        doNotLayer={true}
      >
        {results}
        {status}
      </Callout>
```

unindented: Is that what you meant?

config/jest/config.json
```diff
@@ -38,10 +38,10 @@
   ],
   "coverageThreshold": {
     "global": {
-      "statements": 95,
-      "branches": 95,
-      "functions": 95,
-      "lines": 95
+      "statements": 100,
```
swese44: 👍

src/components/SuggestionsList/SuggestionsList.md
richardtagger: yeah - just need to workout how to give it a non-string ref for the demo then

richardtagger: (your example doesnt work cos the ref cb wont resolve in time to pass it to the sibling)

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -0,0 +1,158 @@
...
+   */
+  groupedItems?: SuggestionItemGroupProps[];
+  /**
+   * The index of current keyboard selection.
+   */
+  selectedId?: string | number;
+  /**
+   * Called when an item in a group is selected by clicking.
+   */
+  onItemSelected(item: SuggestionItem): void;
```
swese44: is there any value returning the entire SuggestionItem instead of its id? the item object is only presentational (id, name, description, imageUrl), could the outer component ever use the full item?

this seems like a nitpicky question, but in general we're trying to pass around specific values rather than bags of things, and we made the decision with Dropdown to only return the id instead of the full item:
https://github.com/Microsoft/YamUI/blob/master/src/components/Dropdown/Dropdown.tsx#L81

richardtagger: but then the user has to look the data up? I wouldnt characterize it as "a bags of things" because you a) pass by object reference b) this is TypeScript so its well defined whats inside

src/components/SuggestionsList/SuggestionsList.md
unindented: You are right. We can keep it as string until React removes it, and then we fix it?

src/components/SuggestionsList/SuggestionsListItem.css
```diff
@@ -0,0 +1,19 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+@import "src/css/variables/colors.css";
+
+.y-suggestions-list-item {
+  cursor: pointer;
+  padding: 4px 12px 8px;
```
swese44: if these paddings need to change with a browser's default text size to maintain vertical rhythm then they should be rem instead of px. maybe these are fine as-is, up to you

src/components/SuggestionsList/SuggestionsListItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const className = search.toLowerCase() === item.toLowerCase() ? highlightedClass : undefined;
+    return <span key={index} className={className}>{item}</span>;
+  });
+};
+
+export default class SuggestionsListItem extends React.PureComponent<SuggestionsListItemProps, {}> {
+  public render() {
+    const { isSelected, isHovered, name, searchText, imageUrl, description } = this.props;
+    const avatar = (
+      <Avatar
```
swese44: this assumes results are all users which display as round avatar in a MediaObject. groups need `borderType: SOFT`, other results with an imageUrl would use an `<Image>` component instead of Avatar for square borders. then there are plain text results which have no image, take up less vertical space, and align all the way to the left instead of using MediaObject.

in the spec meeting we were thinking SuggestionListItem could have a template option. the application would own the business logic of mapping a result type to one of the supported templates:

`type PickerItemTemplate = 'text' | 'mediaRound' | 'mediaSoft' | 'mediaSquare';`

src/components/SuggestionsList/SuggestionsListItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+        imageUrl={imageUrl}
+        name={name}
+        size={AvatarSize.SMALL}
+      />);
+    const className = isHovered ? hoveredClass : isSelected ? selectedClass : baseClass;
+    const title = getHighlightedName(name, searchText);
+    return (
+      <div
+        onMouseDown={this.onMouseDown}
+        onMouseOver={this.onMouseOver}
```
swese44: i think we want onMouseEnter to avoid repeated calls as you mouse over nested elements

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -0,0 +1,158 @@
...
+  groupedItems: SuggestionItemGroupProps[];
+}
+
+export interface SuggestionItemGroupProps {
+  title: string;
+  items: SuggestionItem[];
+}
+
+export interface SuggestionsListState {
+  hoveredId: string | number | null;
```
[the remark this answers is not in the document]
richardtagger: keyboard is selectedId and mouse hover is hoveredId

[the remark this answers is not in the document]
unindented: I didn't follow that. This piece of state is just for the hover? If we're getting the `selectedId` as a prop from our parent, why do we need to keep track of `hoveredId` as local state?

src/components/SuggestionsList/SuggestionsList.md
richardtagger: i added a wrapper with a trigger button - can you take a look at CalloutTrigger? Having them render by default was causing jump around (cos of the auto repositioning magic of Callout). It's better to have them toggle on and off cos that forces rerender and repositioning
screen shot 2017-12-14 at 11 47 33 am

src/components/SuggestionsList/SuggestionsList.tsx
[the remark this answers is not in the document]
richardtagger: keyboard and hovering are distinct - if you see the demo they're different colors. It's the idiom Modern Interfaces use apparently.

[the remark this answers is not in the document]
unindented: Aaah understood.

src/demos/CalloutTrigger/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+    this.target = ref;
+  }
+
+  private onClick = () => {
+    this.setState(({ showing }) => ({ showing: !showing }));
+  }
+
+  public render() {
+    const content = this.state.showing
+    ? React.cloneElement(this.props.children as any, { target: this.target }) : null;
```
unindented: Is this the best that Prettier could do?

richardtagger: we dont have Prettier configured for this project do we?

unindented: Also, by doing `React.cloneElement` you're assuming that you only have a single child right? You may want to call `React.Children.only` https://reactjs.org/docs/react-api.html#reactchildrenonly

unindented: I have Prettier configured in my editor, so I use it everywhere.

Feel free to copy the `.prettierrc` file to this repo though.

src/components/SuggestionsList/SuggestionsList.tsx
unindented: Whatever we do, we should be consistent across components. If `Dropdown` is already doing things one way, either we do the same, or we change both.

I'm fine with just passing the `id`. It's similar to how a plain `select` would behave, where you'd get the `value` of the selected `option`.

swese44: the user always has to look data up because this item is only the visual data (id, name, description, imageUrl), it's not the actual data object managed by the application. in practice, your code will still use selectedItem.id to look up the actual data

src/components/SuggestionsList/SuggestionsListItem.tsx
richardtagger: yup - i'm working along the critical path of my project. In the future when other types of results can be rendered this component can be updated

adrianmg:
I know I'm late to the party and this is already merged but I just wanted to say that this looks awesome. Thanks for taking care of this guys 👍

PR #146, opened 2017-12-12
Rename Callout to Hovercard

swese44:
I'll follow this with another change to extract our Callout component from Hovercard.

Lots of changes but just a rename. Not sure why unrelated visual diff image files are getting updated, there aren't actually visual changes to them.

## Pull request checklist


unindented:
Those reference files you've added... Are they related to examples you renamed/deleted? If so, you should delete them from your folder too, and remove them from this PR.

swese44:
@unindented do we know why the reference images are changing without visual changes? does puppeteer add any file metadata which may change?

swese44:
would it be better to leave the changed images in this PR so they don't change in other upcoming PRs?

swese44:
oh nevermind, they've resolved themselves after syncing from master. only Callout/Hovercard diffs are left

unindented:
I have no idea why Git didn't pick it up as a move. You could try a `git mv` of the old reference files to the new names, and see if that causes any diffs?

swese44:
all good, resolved after syncing from master

adrianmg:
LGTM
image

PR #147, opened 2017-12-12
Extract Callout from Hovercard

swese44:
Couldn't find any reason to limit Fabric Callout's API as this is pretty much a utility component within YamUI, so I just exposed the Fabric component directly. There's still a little value in this wrapper since the CSS override is managed at this entrypoint instead of in Hovercard.

## Pull request checklist

### DIFF
--- a/src/components/Callout/Callout.md
+++ b/src/components/Callout/Callout.md
@@ -1,21 +1,40 @@
 ### Notes for use
 
-Callout is simply exported from Fabric, with some minor YamUI styling. It is meant to be used as a utility component by other YamUI components like Hovercard and Tooltip, you probably don't need to use it directly in your application UIs.
-https://developer.microsoft.com/en-us/fabric#/components/callout
+Callout is simply [exported from Fabric](https://developer.microsoft.com/en-us/fabric#/components/callout), with some minor YamUI styling. It is meant to be used as a utility component by other YamUI components like Hovercard and Tooltip, you probably don't need to use it directly in your application UIs.
+
 
 ### Examples
 
 Basic usage:
 
 ```js
-const visible = true;
+class MyComponent extends React.Component {
+  constructor() {
+    super();
 
-<div>
-  <span className="myTarget">Target</span>
-  {visible && (
-    <Callout target=".myTarget">
-      This message is displayed in the Callout popover!
-    </Callout>
-  )}
-</div>
+    this.state = {
+      visible: false,
+    };
+  }
+
+  render() {
+    const callout = this.state.visible && (
+      <Callout target={this.triggerElement}>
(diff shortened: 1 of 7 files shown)


src/components/Callout/Callout.md
```diff
@@ -0,0 +1,21 @@
+### Notes for use
+
+Callout is simply exported from Fabric, with some minor YamUI styling. It is meant to be used as a utility component by other YamUI components like Hovercard and Tooltip, you probably don't need to use it directly in your application UIs.
+https://developer.microsoft.com/en-us/fabric#/components/callout
```
unindented: I'd wrap the URL in `<>`, or turn some of the words in the paragraph into a link with `[]`. GitHub is doing its own parsing of URLs and turning them into links, but it's not standard Markdown behavior.

src/components/Callout/Callout.md
```diff
@@ -0,0 +1,21 @@
...
+
+Basic usage:
+
+```js
+const visible = true;
+
+<div>
+  <span className="myTarget">Target</span>
+  {visible && (
+    <Callout target=".myTarget">
```
unindented: String refs are getting removed soonish: https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

I wouldn't use them in an example, cause people will copy it.

src/components/Callout/Callout.md
swese44: done

src/components/Callout/Callout.md
swese44: done

src/components/Callout/Callout.md
```diff
@@ -0,0 +1,40 @@
...
+  render() {
+    const callout = this.state.visible && (
+      <Callout target={this.triggerElement}>
+        <div style={{ padding: '10px' }}>This message is displayed in the Callout popover!</div>
+      </Callout>
+    );
+
+    return (
+      <span>
+        <span ref={node => (this.triggerElement = node)} onClick={() => this.setState({ visible: !this.state.visible })}>
```
richardtagger: should this use prevState as the new state relies on the old state
```js
onClick={(prevState) => ({ visible: !prevState.visible })
```

richardtagger: oh this is just an example, ok

src/components/Hovercard/Hovercard.test.tsx
```diff
@@ -182,7 +182,7 @@ describe('<Hovercard />', () => {
 
       it('the Hovercard closes after a delay', () => {
         jest.runTimersToTime(500);
-        expect(component.find(FabricCallout).length).toBe(0);
```
richardtagger: I'd delete all these tests. expect(component).toMatchSnapshot(); means we dont need to be poking around like this

PR #150, opened 2017-12-13
[cleanup] Use snapshots to test rendered output

richardtagger:
We should just be using snapshots to test the rendered output rather than flooding our files with unnecessary prop checks everywhere

## Pull request checklist

### DIFF
--- a/src/components/Text/Text.test.tsx
+++ b/src/components/Text/Text.test.tsx
@@ -12,14 +12,6 @@ describe('<Text />', () => {
       component = shallow(<Text>test content</Text>);
     });
 
-    it('renders its given content', () => {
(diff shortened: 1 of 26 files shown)


unindented:
@swese44 knows my preference for snapshots. I'll let him comment.

src/components/Text/Text.test.tsx
```diff
@@ -112,10 +72,6 @@ describe('<Text />', () => {
       component = shallow(<Text color={TextColor.SECONDARY}>test content</Text>);
     });
 
-    it('renders the correct color className', () => {
```
swese44: and here, "with color secondary"

richardtagger:
Done

PR #158, opened 2017-12-14
Add VSCode Jest Debug script

richardtagger:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Jest Tests",
+      "program": "${workspaceRoot}/node_modules/.bin/jest",
+      "args": [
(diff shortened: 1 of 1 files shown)


richardtagger:
@unindented

.gitignore
```diff
@@ -3,6 +3,7 @@
 /docs
 /node_modules
 /config/travis/deploy.key
-.*
 *.diff.png
 *.new.png
+.DS_Store
+/.jest
```
unindented: Why not ignore all files and folders starting with dot?

richardtagger: Cos we want the .vscode/ folder? Looking at YamJS' we manually list them

unindented: `git add .vscode -f`, and keep the `.gitignore` as it was.

unindented: Things like `.DS_Store` shouldn't live in a project's `.gitignore`. They should live in each person's global `.gitignore`. YamJS is not a good guide for anything.

richardtagger:
ok done

PR #159, opened 2017-12-14
 Remove export of all icons at index level

unindented:
We've seen that Webpack v3 is not tree-shaking the icons, so we'll force consumers to import them one by one for now, until Webpack v4 comes out.

So no more of this:

```js
import { Add } from 'yamui/dist/components/Icon';
```

And a lot of this:

```js
import Add from 'yamui/dist/components/Icon/icons/Add';
```


unindented:
@swese44 @richardtagger please review.

unindented:
@swese44 done

PR #160, opened 2017-12-14
Remove classnames dependency, use es6 modules

swese44:
## Pull request checklist


src/util/classNames.ts
```diff
@@ -0,0 +1,10 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+export type ClassName = string | undefined;
+
+export const join = (classNames: ClassName[]): string => {
+  if (!classNames.length) {
```
unindented: I'm not a fan of type coercions, I'd have explicitly compared with `=== 0`.

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "module": "commonjs",
+    "target": "es6",
```
unindented: Why also change the target? If you change the target, then consumers will have to transpile. If you leave the target as `es5` and change the module type to `es6`, webpack will be able to do the right thing.

swese44:
udpated

src/components/FakeLink/FakeLink.tsx
```diff
@@ -19,6 +19,6 @@ export default class FakeLink extends React.Component<FakeLinkProps, {}> {
   render() {
     const { className, children } = this.props;
 
-    return <span className={classNames('y-fakeLink', className)}>{children}</span>;
+    return <span className={join(['y-fakeLink', className])}>{children}</span>;
```
richardtagger: why not just use Array.prototype's join?

richardtagger: ah I see, it is underneath. k

swese44: this removes empty/undefined items so we don't have to check for the existence of `className` in each component

PR #166, opened 2017-12-16
Add missing functionality to Dropdown

### LINKED ISSUE #164: Dropdown missing option templates
icon, bolded text, text

### LINKED ISSUE #165: Dropdown missing dividers and headers

swese44:
- Adds section header option type
- Adds divider option type
- Adds icon and label support in option templates
- Prevents Dropdown from closing on scroll
- Allows Dropdown to open above instead of below if there isn't room below
- Maps Fabric's IDropdownOption to our own DropdownOptions interface to keep our API surface as simple as possible. Change is backward-compatible with our previous implementation.

Fixes #164
Fixes #165

It would be nice if we could add scripting for visual diffs to open hovercards, dropdowns, etc.

screen shot 2017-12-15 at 9 36 59 pm

## Pull request checklist


src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -50,6 +51,80 @@ describe('<Dropdown />', () => {
...
+      );
+      fullComponent.find('.ms-Dropdown').simulate('click');
+    });
+
+    it('matches its snapshot', () => {
+      expect(fullComponent.find('.y-dropdown--item')).toMatchSnapshot();
+    });
+  });
+
+  xdescribe('with divider', () => {
```
unindented: `xdescribe`

src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -50,6 +51,80 @@ describe('<Dropdown />', () => {
     });
   });
 
+  xdescribe('with section header', () => {
```
unindented: `xdescribe`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -2,14 +2,48 @@
...
+
+  /**
+   * The visible option text.
+   */
+  text: string;
+
+  /**
+   * Unique identifier for the given option.
+   */
+  key: DropdownOptionKey;
```
unindented: I'd put mandatory attributes above optional ones, as they're probably more important.

src/components/Dropdown/Dropdown.test.tsx
swese44: ah, meant to delete those actually

swese44:
@mamuso thanks for the updates! i'm readjusting the new CSS, a few quick questions:

1. you updated the dropdown options to use $textColor__secondary, but the title (the selected option visible when the dropdown is closed) still uses $textColor. is this intentional, or should it also use $textColor__secondary?
2. the zeplin mockups show some examples using $textColor for all options + title, and some using $textColor__secondary for all options + title. is that intentional, or is $textColor__secondary the correct one?

thanks!

swese44:
(fyi i'm replacing the relative positioning so we don't get into situations like overflow ellipsis being a pixel too low)
screen shot 2017-12-18 at 5 28 58 pm

mamuso:
@swese44 Using two different colors on the list of options is intentional. I'll go back and update the mocks.

swese44:
@mamuso ok so title is primary text color, options within the dropdown are secondary text color?

also, should long options ellipsis instead of wrapping? currently they wrap:
screen shot 2017-12-18 at 5 31 42 pm

mamuso:
@swese44 yes, options within the dropdown are secondary.

On wrapping vs ellipsis, I think we need to wrap (probably touching the top/bottom padding a bit if possible). If we end up with that issue, you want to be able to read all the text available to you there.

swese44:
i think we're in pretty good shape. @mamuso can you take another pass when you get a chance? thanks!

screen shot 2017-12-20 at 1 07 33 am
screen shot 2017-12-20 at 1 07 04 am

Issue #7, opened 2017-07-12
Add "type" prop for button


swese44:
const ButtonType = 'button' | 'submit' | 'reset';
Should default to "button".

unindented:
Fabric doesn't support this, right? I don't see any issues open in their repo. Should we open one?

swese44:
@unindented Fabric also accepts HTML attribute props, so you could pass in type or other `<button>` attributes and it should pass them along to the DOM

swese44:
(i have not verified this specific case, but it should work)

PR #171, opened 2017-12-18
Add type to button

chrismohr:
Fixes https://github.com/Microsoft/YamUI/issues/7

## Pull request checklist


src/components/Button/Button.md
```diff
@@ -4,6 +4,8 @@ While buttons can technically be used to navigate a user to another part of the
 
 Also, use concise, specific, self-explanatory text, usually a single word. They should include a verb. If acting on something, their text should include the noun they're acting on, e.g. *"Create Group"*.
 
+Use type='submit' for submitting forms.  Doing this will trigger the onClick action when the user presses the enter key within certain form fields.
```
unindented: Wrap things that should look like code in backticks (e.g. `type="submit"`), so that they get rendered properly in Markdown.

unindented: There's two spaces before `Doing`. (My OCD kills me.)

src/components/Button/Button.test.tsx
```diff
@@ -1,7 +1,8 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 import * as React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
-import Button, { ButtonProps, ButtonColor, ButtonStatus, ButtonIconPosition, ButtonSize } from '.';
+import Button,
+  { ButtonProps, ButtonColor, ButtonStatus, ButtonIconPosition, ButtonSize, ButtonType } from '.';
```
unindented: Do you use Prettier for code formatting?

src/components/Button/Button.tsx
```diff
@@ -149,13 +156,15 @@ export default class Button extends React.Component<ButtonProps, {}> {
     const href = (this.props as LinkButtonProps).href;
     const status = (this.props as RegularButtonProps).status;
     const isDisabled = status === ButtonStatus.DISABLED || status === ButtonStatus.LOADING;
+    const type = this.props.type || ButtonType.BUTTON;
```
unindented: Why here instead of in `defaultProps`?

src/components/Button/Button.tsx
```diff
@@ -149,13 +156,15 @@ export default class Button extends React.Component<ButtonProps, {}> {
...
 
     return (
       <BaseButton
         ariaLabel={ariaLabel}
         className={this.getClasses()}
         disabled={isDisabled}
         href={href}
+        type={href ? '' : type}
```
unindented: We may not want to allow `type` when the `Button` is a link. @swese44 should we deal with this attribute in `RegularButtonProps` instead?

src/components/Button/Button.md
chrismohr: Haha... old habits die hard.  I looked for some support from google, but all I found was strongly worded articles that say 2 spaces is definitely wrong.  I'll fix it. :-)

src/components/Button/Button.test.tsx
chrismohr: I've upped the line limit to 120 to be consistent with yamjs. :-)

chrismohr: (and now have prettier)

src/components/Button/Button.tsx
chrismohr: Oversight.  This is fixed now.

src/components/Button/Button.tsx
chrismohr: I've now done `type?: undefined;` on `LinkButtonProps`.  Does that work?

unindented: I think we've done `type?: void` for other attributes in `Button`.

PR #172, opened 2017-12-19
Introduce TextArea

chrismohr:
Introduce TextArea Component

Additionally, add onChange prop to textfield.

## Pull request checklist


chrismohr:
@mamuso The way I did this in https://github.com/Microsoft/YamUI/pull/140 was by overriding the fabric theme colors. (However I had previously missed the error color, which I'll now add.)

Now the active/focus color is #6c98d9, and the error color is #d1423b.

What do we want for hover?  Fabric uses #212121.  Do we even want a hover style? Or should I just force the style to stay the same?

mamuso:
@chrismohr we don't have a hover style. You got focus and error right, but our regular border style is #DDE0E6. Could we update that?

mamuso:
@chrismohr my bad! the default border color looks good now

chrismohr:
@mamuso My most recent commits should have changed the regular border style to `#DDE0E6`.  When I use the digital color meter on the reference images in this PR, it seems correct.  Does it still look off to you?

src/components/TextArea/TextArea.css
```diff
@@ -0,0 +1,47 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+@import "src/css/variables/colors.css";
+@import "src/css/variables/fonts.css";
+
+// work is in progress to set styling/theme options for all Fabric label instances
```
swese44: CSS comments should be wrapped in `/* comment */`

src/components/TextArea/TextArea.md
```diff
@@ -0,0 +1,42 @@
+### Examples
+
+Basic
+```js { "props": { "data-example": "basic" } }
+<TextArea
+  value="This multiline has fixed height so scrollbars may appear."
```
swese44: can we add lines of content so the scrollbars show in the visual diff (or at least the overflow if Chrome auto-hides scrollbars)?

src/components/TextArea/TextArea.md
```diff
@@ -0,0 +1,42 @@
...
+  rows={3}
+/>
+```
+
+Auto-adjust Height
+```js { "props": { "data-example": "autoAdjustHeight" } }
+<TextArea
+  value="This multiline auto grows."
```
swese44: can we add lines of content so the height is taller for visual diffs?

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,142 @@
...
+   */
+  onMouseLeave?: (event: React.MouseEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
+}
+
+/**
+ * The TextArea component enables a user to type text into an app. It's typically used to capture
+ * a single line of text, but can be configured to capture multiple lines of text. The text
+ * displays on the screen in a simple, uniform format.
+ */
+export default class TextArea extends React.PureComponent<TextAreaProps, {}> {
```
swese44: I would have first extracted our Textfield implementation + styling to BaseTextField, then added the necessary TextArea functionality to BaseTextField

src/components/TextArea/TextArea.css
```diff
@@ -0,0 +1,47 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
```
swese44: let's not duplicate this CSS file. let's get TextField merged so these components can share styles

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,142 @@
...
+  private async: Async;
+  private debouncedOnChange: (newValue: string) => void;
+
+  constructor(props: TextAreaProps) {
+    super();
+    if (props.onChangeDebounceTime && props.onChange) {
```
swese44: let's not duplicate these implementations. after the TextField PR is merged, TextField and TextArea can be simple wrappers around BaseTextField which handles all styling and actual functionality

src/components/TextArea/TextArea.css
chrismohr: agreed

src/components/TextArea/TextArea.md
chrismohr: good thinking

src/components/TextArea/TextArea.tsx
chrismohr: Agree.  That's what I was suggesting here.
https://www.yammer.com/microsoft.com/#/Threads/show?threadId=1002522017

src/components/TextArea/TextArea.tsx
chrismohr: My ruby experience has taught me that taking the path of independently implementing the two (Ideally 3) like things, and then extracting the common parts usually leads to better abstractions. I guess we'll see. :-)

mamuso:
Is this stuck for some reason?

unindented:
Travis is complaining, saying that the diffs for the TextArea screenshots are different.

unindented:
Also, I think @chrismohr had a pending refactor that he hasn't pushed yet.

mamuso:
Oh. I thought I pushed the new screens. Will do once I get to a computer

chrismohr:
@mamuso @unindented Yes, I'm still working on this. I intend to pick it up again early next week.

src/components/TextField/TextField.test.tsx
```diff
@@ -1,22 +1,14 @@
...
   let component: ShallowWrapper<TextFieldProps, {}>;
 
   describe('with default options', () => {
     beforeEach(() => {
-      component = shallow(<TextField />);
-    });
-
-    it('has its correct base class', () => {
-      expect(component.hasClass('y-text-field')).toBe(true);
+      component = shallow(<TextField />).dive().dive();
```
chrismohr: `.dive().dive()` let's us snapshot what we're passing to the fabric component.

chrismohr:
@swese44 @unindented can you review please?

swese44:
quick question as i'm getting started reviewing. we had discussed extracting a BaseTextField component to manage styling and behavior for both cases, which Textfield and Textarea would each compose and restrict the props as necessary. is there a reason you took this other approach?

src/components/TextArea/TextArea.css
```diff
@@ -0,0 +1,7 @@
...
+@import "src/css/base-text-field.css";
```
swese44: we're essentially still duplicating the BaseTextField CSS file by importing its contents into both TextArea and TextField. if we're taking this approach instead of composition it is probably better for TextArea.tsx and TextField.tsx to import 'base-text-field.css' so webpack can include it once.

swese44:
looks like TextArea_withDescription_desktop.png should be deleted?

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,54 @@
...
+        resizable={false}
+        onChanged={this.props.unifiedOnChange}
+        {...getBaseTextFieldProps(this.props)}
+      />
+    );
+  }
+
+  getClasses() {
+    const { className } = this.props;
+    const classes = ['y-base-text-field y-text-area'];
```
swese44: any reason these are a single string element in the array?

chrismohr:
@swese44 Looks to me like `TextArea_withDescription_desktop.png` should still be there, since the `TextArea.md` still has a `withDescription` example.  Am I missing something?

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    if (className) {
+      classes.push(className);
+    }
+
+    return classes.join(' ');
```
swese44: if you use the join utility it handles empty values. instead of a getClasses() method you could just inline:

```
className={join(['y-base-text-field', 'y-text-area', className)}
```

src/components/TextField/TextField.tsx
```diff
@@ -152,3 +63,7 @@ export default class TextField extends React.PureComponent<TextFieldProps, {}> {
     return classes.join(' ');
```
swese44: same thing here, we could simplify this with our join utility:

```
className={join(['y-base-text-field' 'y-text-field', className, prefix, suffix])}
```

src/components/TextArea/TextArea.tsx
chrismohr: Nice. I'll try it.

src/components/TextArea/TextArea.md
```diff
@@ -0,0 +1,51 @@
...
+<TextArea
+  value="This multiline has fixed height so scrollbars may appear. This multiline has fixed height so scrollbars may appear. This multiline has fixed height so scrollbars may appear. This multiline has fixed height so scrollbars may appear. This multiline has fixed height so scrollbars may appear. This multiline has fixed height so scrollbars may appear."
+  rows={3}
+/>
+```
+
+With a description
+```js { "props": { "data-example": "withDescription" } }
+<TextArea
+  value="This multiline has fixed height so scrollbars may appear."
```
swese44: ah i was confused on the screenshot, it doesn't look like this text reflects the example. maybe the text could read "This Textarea has a description"

chrismohr: Good call. I'll do that.

src/components/TextArea/TextArea.md
```diff
@@ -0,0 +1,51 @@
...
+  rows={3}
+  description="Small hint goes here"
```
swese44: maybe this could say "This is the Textarea's description"

src/components/TextField/TextField.tsx
chrismohr: I updated the TextArea with your suggestion.  This one isn't quite as simple.  `prefix` and `suffix` don't contain the class names.  I took another look at the implementation, and I'm not seeing much room to simplify without sacrificing readability.  Let me know if you have another suggestion though.

PR #176, opened 2017-12-22
SuggestionsList - create index file

richardtagger:
Need to actually be able to import `components/SuggestionsList`


richardtagger:
@unindented @swese44

unindented:
Coverage went down because of the new file. Maybe tests should be importing the index file? That way we'd also prevent the mistake you made of not providing an index.

richardtagger:
updated

PR #178, opened 2017-12-24
[avatar] Support disabling fade in

richardtagger:
For highly dynamic content e.g. lists, drop-downs etc, Avatar images fading in every render is distracting, this adds a prop to pass through to Persona's imageShouldFadeIn

## Pull request checklist

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -44,12 +44,17 @@ export interface AvatarProps extends BaseComponentProps {
    */
   imageUrl?: string;
 
-
   /**
    * XLARGE: 72px, LARGE: 48px, MEDIUM: 40px, SMALL: 32px, XSMALL: 24px.
    * @default AvatarSize.MEDIUM
    */
   size?: AvatarSize;
+
+  /**
+   * Will hide the image until it has loaded, then fade it in.
+   * @default false
+   */
+  imageShouldFadeIn?: boolean;
 }
 
 /**
@@ -62,7 +67,7 @@ export default class Avatar extends React.Component<AvatarProps, {}> {
   };
 
   render() {
-    const { badgeContent, imageUrl, name, size } = this.props;
+    const { badgeContent, imageUrl, name, size, imageShouldFadeIn } = this.props;
     const personaSize = SizeMap[size as string];
 
     const badge = badgeContent && (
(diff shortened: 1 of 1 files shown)


richardtagger:
@unindented

src/components/Avatar/Avatar.tsx
```diff
@@ -50,6 +50,12 @@ export interface AvatarProps extends BaseComponentProps {
    * @default AvatarSize.MEDIUM
    */
   size?: AvatarSize;
+
+  /**
+   * Decides whether to fade the Avatar image in each render, pass through to Persona
+   * @default false
+   */
+  imageShouldStartVisible?: boolean;
```
unindented: Should we keep the same prop name we used with `Image`? https://github.com/Microsoft/YamUI/blob/9d6a78a3a25241dfb8b38becec8b022bccdee1ca/src/components/Image/Image.tsx#L48-L52

richardtagger: the interesting thing from Persona is it has these both:-

```ts
    /**
     * If true, adds the css class 'is-fadeIn' to the image.
     */
    imageShouldFadeIn?: boolean;
    /**
     * If true, the image starts as visible and is hidden on error. Otherwise, the image is hidden until
     * it is successfully loaded. This disables imageShouldFadeIn.
     * @defaultvalue false
     */
    imageShouldStartVisible?: boolean;
```

I'll change it to use `imageShouldFadeIn`

richardtagger: done

src/components/Avatar/Avatar.tsx
```diff
@@ -59,10 +64,11 @@ export default class Avatar extends React.Component<AvatarProps, {}> {
   static defaultProps: Partial<AvatarProps> = {
     borderType: AvatarBorderType.ROUND,
     size: AvatarSize.MEDIUM,
+    imageShouldFadeIn: true,
```
unindented: Should it default to `true`? I think designers were complaining about the fade-in behavior, so I'd default it to `false`.

richardtagger: ok, sure.

richardtagger: updated

swese44:
👍

PR #179, opened 2017-12-25
[MediaObject] Use ellipsis for text

richardtagger:
Text contents should not flow to next line if too long for width of container

## Pull request checklist


richardtagger:
@unindented

unindented:
Should the same apply for meta and extra?

richardtagger:
done

swese44:
was this requested by UX? please add long examples with visual diffs, UX should sign off on these visual changes

richardtagger:
nope - just looks weird.  I asumed this content was always intended to be a single line. e.g. 
screen shot 2017-12-25 at 12 07 35 am
screen shot 2017-12-24 at 4 00 13 pm

swese44:
there are examples of MediaObject here which don't use ellipsis

swese44:
ya please check with UX before merging, i'm not sure what the visual requirements are. it may be better to default to ellipsis text for title and/or subtexts, but still allow a prop like `allowWrappedText`

richardtagger:
are the recent activities meant to be MediaObjects too? so this should be a prop?

swese44:
i don't know. @mamuso or @adrianmg could you provide UX requirements on if/when MediaObject title or subtexts should wrap or use ellipsis? thanks!

either way let's add examples with visual diffs to make sure we don't accidentally break this in the future

richardtagger:
I've added an example - if it's determined we need to support wrapping lines down the road that's an easy prop to add. Currently nothing consumes MediaObject - I'll be the first and most use cases from Zeplin seem to be single line.

mamuso:
If we are not repurposing media objects for other cases, then they should handle the ellipsis

swese44:
sweet thanks @richardtagger. we can always add a prop if we need to extend MediaObject and allow wrapping

PR #180, opened 2017-12-26
[SuggestionsList] Change mouseOut to mouseLeave

richardtagger:
Mouse out causes flickering

### DIFF
--- a/src/components/SuggestionsList/SuggestionsList.test.tsx
+++ b/src/components/SuggestionsList/SuggestionsList.test.tsx
@@ -99,7 +99,7 @@ describe('<SuggestionsList />', () => {
             rendered
               .find('ul')
               .at(1)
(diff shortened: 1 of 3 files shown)


unindented:
Why does it produce flickering? Can you detail? I think @swese44 was originally concerned about `onMouseOver` because of the amount of events we'd generate. Can you record the React Dev Tools with *Highlight Updates* enabled, so that we see what we're re-rendering atm?

richardtagger:
YamJS branchname `editor-container` has it for you to have a look

unindented:
Could you record with Quicktime and post on Yammer? Or create a GIF and post here? I'm a bit swamped with PRs.

swese44:
sorry i was actually not working on my vacation. this change is correct, onMouseLeave matches the onMouseEnter event which we're also using

PR #181, opened 2017-12-26
[SuggestionsList] props.target to match Callout

richardtagger:
Pass through prop type def should match underlying Component's

## Pull request checklist

### DIFF
--- a/src/components/SuggestionsList/SuggestionsList.tsx
+++ b/src/components/SuggestionsList/SuggestionsList.tsx
@@ -8,12 +8,13 @@ import Text, { TextSize, TextColor } from '../Text';
 import Callout, { DirectionalHint } from '../Callout';
 import Block, { GutterSize } from '../Block';
 import './SuggestionsList.css';
+import { IPoint } from 'office-ui-fabric-react/lib/utilities/positioning';
 
 export interface SuggestionsListProps extends NestableBaseComponentProps {
   /**
    * The target that the Callout should try to position itself based on.
    */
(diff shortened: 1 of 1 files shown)


richardtagger:
@swese44 @unindented

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -8,12 +8,13 @@ import Text, { TextSize, TextColor } from '../Text';
...
 import Block, { GutterSize } from '../Block';
 import './SuggestionsList.css';
+import { IPoint } from 'office-ui-fabric-react/lib/utilities/positioning';
 
 export interface SuggestionsListProps extends NestableBaseComponentProps {
   /**
    * The target that the Callout should try to position itself based on.
    */
-  target: HTMLElement;
+  target?:  HTMLElement | string | MouseEvent | IPoint | null;
```
unindented: Why is it optional and `null`? Where will it render when it's `undefined` or `null`?

richardtagger: Just aligned it to Fabrics. The only one I intend to use is IPoint

unindented: Right, but what's the effect of `undefined` and `null`? If it doesn't make sense in our scenario, no sense in aligning.

richardtagger: ok done

PR #191, opened 2018-01-05
add allowTextWrap prop to MediaObject

jessemartin:
Previously, MediaObject truncates content text with an ellipsis by default. This behavior is not always desired. Sometimes content text is important enough to wrap instead of truncate.

This PR adds an `allowTextWrap` option to MediaObject. If provided, all overflowing text (titleContent, metadataContent and extraContent) will wrap to the next line. Otherwise, overflowing text will be truncated with an ellipsis.

This PR also changes `Block` to use [`overflow-wrap: break-word`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap). As a result, the IconsLibrary in [Icons.md](src/components/Icon/docs/Icons.md) required modification to prevent a long icon name from throwing off the icon grid layout.

## Pull request checklist


jessemartin:
@swese44 @unindented @mrfunkycold

unindented:
Should we have a visual example of a `MediaObject` with wrapping text? That way it'll be more difficult to introduce regressions.

src/components/MediaObject/MediaObject.tsx
```diff
@@ -41,6 +41,11 @@ export interface MediaObjectProps extends NestableBaseComponentProps {
    * content like buttons or other elements, pass those in as children instead of here.
    */
   extraContent?: React.ReactNode;
+
+  /**
+   * Limits text content to a single line, hiding additional text with an ellipsis.
+   */
+  ellipsis?: boolean;
```
swese44: This will break existing implementations which expect the ellipsis by default. Most use cases will actually need the text wrapping, would it make more sense to change this to `allowTextWrap` and keep the default as ellipsed?

swese44: the rest of the MediaObject sub-components could remain as-is, they are not exposed publicly

src/components/MediaObject/MediaObject.md
```diff
@@ -102,6 +102,64 @@ const avatar = (size) => (
 </div>
 ```
 
+With ellipsis:
+
+```js { "props": { "data-example": "with ellipsis" } }
```
swese44: if we change `ellipsis` to `allowTextWrap` then we should also update this test name to "with allow text wrap"

src/components/MediaObject/MediaObject.md
```diff
@@ -102,6 +102,64 @@ const avatar = (size) => (
...
+const name = 'Hubert Blaine Rindfleischetikettierungsüberwachungsaufgabenübertragungsgesetz';
+const title = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt';
+const status = 'Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque';
+
+const avatar = (size) => (
+  <Avatar imageUrl={user.imageUrl} name={user.name} size={size} />
+);
+
+<div>
+  <MediaObject
```
swese44: we probably don't need to show the example at every size, just one size should be fine to show the text wrapping

swese44:
the visual diffs are probably failing because we've changed the default ellipsis behavior of MediaObject

src/components/MediaObject/MediaObject.tsx
jessemartin: @swese44 I chatted with @unindented about this briefly. `ellipsis` is being used already in other components. We felt that consistency with those components was more important.

Do you think we should go with allowTextWrap instead of updating the existing usages to include `ellipsis={true}`?

jessemartin:
@swese44 @unindented Updated visual diffs are on the way!

src/components/MediaObject/MediaObject.tsx
unindented: Yeah, I feel that, if `Block` and `Text` already have an attribute `ellipsis`, and it is off by default, then adding an opposite attribute here would make things harder to understand.

swese44: Block has `ellipsis`, Text has `maxWidth` which implicitly adds the ellipsis.

somebody deleted the UI Domain tenants from the group sidebar, but we had tenants like "provide smart defaults" and "make it hard to do the wrong thing, easy to do the right thing". if we default ellipsis to false here than most implementations will need to specify `ellipsis={true}`. we could default MediaObject's ellipsis to true, but that's the opposite of Block's default and might not be intuitive.

i'm ok with either prop name but i prefer `allowTextWrap`, and MediaObject should ellipse text by default

swese44: if you guys feel strongly then we can go with the default value being false. if so we should publish a major version and update all existing instances in YamJS with `ellipsis={true}`

unindented: Understood. My only worry then is referring to the same concept with three different names: `ellipsis`, `maxWidth` and `allowTextWrap`.

swese44: ya it's a tradeoff. i still prefer defaulting to the main use case of ellipsed text and using `allowTextWrap` as the override. if you guys prefer defaulting to wrapped text with `ellipsis` then be sure to follow this up with a major version bump 5.0.0 and update instances in YamJS before anybody else bumps us to 5.x

jessemartin: Okay, it sounds like the most common use case should use truncation via ellipsis, so I'll diverge from the `ellipsis` prop here. `allowTextWrap` incoming shortly!

unindented: Can we rename `ellipsis` to something that has `textWrap` in the name then? `disallowTextWrap`? I still think naming the same concept with three different names in three components is a Bad Idea™.

jessemartin:
@unindented @swese44 I've migrated to using an `allowTextWrap` prop. This is now ready for review.

Side question - does this functionality require a version bump? It won't break existing usages.

unindented:
If it doesn't break anything, then version bump will be either `minor` or `patch`.

adrianmg:
Is this expected? 
image

swese44:
might need that on the individual `title`, `metadata`, `extra` sub-components

jessemartin:
@adrianmg Text wrapping has been addressed:
screen shot 2018-01-10 at 15 15 58

PR #199, opened 2018-01-10
Issue 197: Colorized block

### LINKED ISSUE #197: Add color to Block
We have cases where Text is nested in Block just for color. If Block has color then these would not be necessary. It would also help people get correct line-heights easier since Block will always have the correct line-height. It will also let us remove some override CSS in MediaObject.

jessemartin:
Adds `textColor` prop to Block (fixes #197). This allows removal of nested Text components when used with Block solely to specify color.

## Pull request checklist


jessemartin:
@swese44

PR #209, opened 2018-01-17
[Avatar] Fix default prop

richardtagger:
It appears a falsey ‘undefined’ value isn’t enough. Explicitly setting false is required

## Pull request checklist


richardtagger:
@swese44

PR #222, opened 2018-01-24
[SuggestionList] Stop component repositioning itself

richardtagger:
This will keep the list position locked to below instead of it glitching around looking janky


richardtagger:
@swese44 @unindented

PR #224, opened 2018-01-26
Release version 4.3.3

richardtagger:
* SuggestionsList will now always appear below target


richardtagger:
@unindented

swese44:
published

richardtagger:
thanks!

PR #228, opened 2018-01-29
[WIP] PreviewCard component and related changes

### LINKED ISSUE #220: Box border pushes content in 1px
image

The left one is implementation, right is the design spec.

The total size is wrong (it should be 6px less), probably due to the fact that we're taking into account the border width as part of the total size of the component (among other small issues). 

Also, the margin between the end of the media object and the progress indicator should probably be 5px smaller (or maybe smaller, depends on how we handle the border thing).

--

I'd also say that the 'MediaObject child' is probably wrong too for the same exact reason. 90px height doesn't sound like a round number compared to the 88px that should have if we weren't counting the borders as part of the total size.

If this helps: we're *don't count* in boxed content the width of the border as a parameter that affects the total size. So for instance, if we say a box of 64px height, that would include the border within that box, not an outer border.

swese44:
- Adds PreviewCard
- Adds abstracted EditableText component
- Fixes CSS/markup on Box, adds onClick handler. Fixes #220
- Fixes a height/layout bug in ProgressIndicator
- Adds new metadata color to Spinner
- Sets Spinner's default layout to inline, with `centered` as an option (breaking change)
- Sets underlined Textfield to the correct height
- Adds .focus() public method to TextField
- Surfaces some new Fabric TextField props for new use cases
- Adds tslint rules to match our other codebase (Prettier was saving conflicting formats)

Note that we'll probably need to wait to implement PreviewCard's filetype illustrations until we bring illustrations into YamUI.

## Pull request checklist
* [ ] Support className on PreviewCard
* [ ] Fix onClick handler to only trigger `this.props.onClick` if not in edit mode
* [ ] Fix all new linting issues in the repo
* [ ] Unit tests for PreviewCard
* [ ] Unit tests for EditableText
* [ ] Fix icon size in EditableText. Or maybe better to document that the component will render appropriately based on the nearest Block or Text component with textSize, and just wrap the .md example in `<Text size={TextSize.XSMALL}>`
* [ ] Light gray color for EditableText content (icon + text), darker gray on hover
* [ ] EditableText should support className
* [ ] Readme examples for EditableText
* [ ] Unit tests for new Spinner/Box/TextField functionalities
* [ ] Readme examples for new Spinner/Box/TextField functionalities


swese44:
should have fixed prettier+linting issues in its own PR...

swese44:
replacement PR: https://github.com/Microsoft/YamUI/pull/287

PR #233, opened 2018-02-01
VisibilityObserver and KeyboardNavigationObserver components

swese44:
**VisibilityObserver**
This abstracts ReactIntersectionObserver behind a simpler API. Requires the client application to include the IntersectionObserver polyfill for older browsers, i.e. IE11. Also replaces custom example wrapper with the new component.

**KeyboardNavigationObserver**
Listens to mouse events and navigational keyboard events to conditionally render content and callbacks.

These 2 components can be used together to build accessible infinite scroll features.

There is a PropTypes console warning from `react-intersection-observer`, hopefully they'll fix that soon :(

## Pull request checklist


src/components/VisibleRender/VisibleRender.md
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <div style={{ backgroundColor: '#f98985', transition: 'background-color 1s' }}>
+      <Block padding={GutterSize.XXLARGE} textSize={TextSize.LARGE}>Out of view</Block>
+    </div>
+  );
+};
+
+<VisibleRender 
+  whenInView={renderInView}
+  whenOutOfView={renderOutOfView}
```
unindented: For `ResponsiveRender` in the new client, you went with prop names `renderAbove` and `renderBelow`. Should we align with these and make these `renderInView` and `renderOutOfView`, or make the other ones `whenAbove` and `whenBelow`?

Also, should we bring `ResponsiveRender` into YamUI?

src/components/VisibleRender/VisibleRender.md
```diff
@@ -0,0 +1,30 @@
...
+  onEnter={action('scrolled into view')}
+  onLeave={action('scrolled out of view')}
```
unindented: Should these have `View` in the name to be clearer, and to create a connection with the other two props? `onEnterView` and `onLeaveView`? Or `onInView` and `onOutOfView`, but they sound worse...

src/components/VisibleRender/VisibleRender.md
swese44: good question. ResponsiveRender ties `react-media` to our breakpoints, do we want those breakpoints directly in YamUI? it would actually make it easier keep consistency between clients. let's do this in another PR

swese44: and yes i'll rename these render props to include "render"

src/components/VisibleRender/VisibleRender.md
swese44: will rename

src/components/VisibleRender/VisibleRender.tsx
```diff
@@ -0,0 +1,82 @@
...
+   */
+  onLeave?: () => void;
+
+  /**
+   * Render prop to return child content when the component is visible.
+   */
+  whenInView?: () => React.ReactNode;
+
+  /**
+   * Render prop to return child content when the component is visible.
```
jessemartin: Render prop to return child content when the component is not visible.

PR #237, opened 2018-02-03
Suggestions list fixes

richardtagger:
* SuggestionListItem uses pseudo-class :hover to highlight rather than involving JavaScript
* SuggestionListItem highlights with bold rather than underline at the request of design
* SuggestionList is now allowed to flow off-screen, working around Callout's default clipping behavior.
* Disables animation

## Pull request checklist


richardtagger:
@adrianmg @swese44 @unindented

src/components/SuggestionsList/SuggestionsList.css
```diff
@@ -1,6 +1,10 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 @import "src/css/variables/colors.css";
 
+.y-suggestions-list.ms-Callout {
+  animation-name: none;
+}
```
richardtagger: @adrianmg here's a chance to change the box-shadow/border-radious stuff. The default looks like it matches Zeplin but I think you wanted to change it?

current settings:-
```cs
border-radius: 1px;
box-shadow: 0 0 12px 0 rgba(0, 0, 0, 0.2);
```

adrianmg: The border-radius might be applied to the container but is probably not being displayed properly due to the border-radius value of the first and last row.

Attaching the image showing this…

image

adrianmg: Left is wrong, right is correct

image

richardtagger: ah ok, taking a look

PR #239, opened 2018-02-05
[Callout] Apply border radius

richardtagger:
Apply border-radius to Callout.

## Pull request checklist


richardtagger:
@adrianmg this looks like its working for Hovercard but there was no changes to Dropdown. I'm not sure what I need to change for that, I was having a play in devtools but couldnt really see what needs to change - there's already a border there that's 1px grey and rounded

adrianmg:
@richardtagger maybe due to the same reason it wasn't rendering properly on the other component, these are not rendering the corner properly. It's literally looking the same way. Maybe it's due the fact the mediaobjects or other parts are rendering the background and not taking the radius into account or something.

PR #241, opened 2018-02-05
Fix keyboardNavigation to run after DOMContentLoaded

richardtagger:
This PR delays initialization of keyboardNavigation until after document.body is ready

## Pull request checklist

### DIFF
--- a/src/util/accessibility/keyboardNavigation.ts
+++ b/src/util/accessibility/keyboardNavigation.ts
@@ -6,7 +6,10 @@ import KeyboardNavigationEventListener, {
 
 export { NavigationMode, NavigationModeCallback };
 
-const navListener = new KeyboardNavigationEventListener(document.body);
+let navListener: KeyboardNavigationEventListener;
+document.addEventListener('DOMContentLoaded', () => {
+  navListener = new KeyboardNavigationEventListener(document.body);
+});
 
 export const getMode = (): NavigationMode => {
   return navListener.getMode();
--- a/src/util/accessibility/keyboardNavigation.ts
+++ b/src/util/accessibility/keyboardNavigation.ts
@@ -7,18 +7,24 @@ import KeyboardNavigationEventListener, {
 export { NavigationMode, NavigationModeCallback };
 
 let navListener: KeyboardNavigationEventListener;
-document.addEventListener('DOMContentLoaded', () => {
-  navListener = new KeyboardNavigationEventListener(document.body);
-});
+
+const ensureInit = () => {
+  if (!navListener) {
+    navListener = new KeyboardNavigationEventListener(document.body);
+  }
+};
 
 export const getMode = (): NavigationMode => {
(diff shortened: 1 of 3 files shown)


unindented:
Can you explain why this was necessary?

richardtagger:
because document.body can be null which the underlying Fabric components dont expect so it was blowing up with an NPE

richardtagger:
i.e. we're running 

```js
navListener = new KeyboardNavigationEventListener(document.body);
```

but document.body might not be ready depending on how the scripts are loaded I guess. Embed widget highlighted this issue

unindented:
I didn't think `document.body` is null before `DOMContentLoaded`. Maybe it's only null if you have the offending script in your head?

src/util/accessibility/keyboardNavigation.ts
```diff
@@ -6,7 +6,10 @@ import KeyboardNavigationEventListener, {
 
 export { NavigationMode, NavigationModeCallback };
 
-const navListener = new KeyboardNavigationEventListener(document.body);
+let navListener: KeyboardNavigationEventListener;
```
unindented: Now the true type of this variable is `KeyboardNavigationEventListener | undefined`.

unindented: (because until `DOMContentLoad` it will be undefined)

unindented: So you'll need to add `if` guards wherever we access the variable. Otherwise shit would blow up if the method gets called before `DOMContentLoad`.

unindented:
This is going to complicate the implementation, as you'll need to add `null/undefined` checks. Is there a way to not add this shit to YamUI?

richardtagger:
yes it's to do with having the script tag in your head. DOMContentLoaded guarantees its all available right? I did test the fix against embed and it resolved it, but are you saying that's incidental?

richardtagger:
Here's the line in Fabric that blows up https://github.com/OfficeDev/office-ui-fabric-react/blob/724a69333e9c1b0833ef6f48c0fd6b49a2bf3a3a/packages/utilities/src/EventGroup.ts#L208

richardtagger:
part of the problem is because target is any'd so undefined is acceptable

swese44:
how about lazy instantiating like this latest commit?

richardtagger:
¯\_(ツ)_/¯ sure. @unindented ?

swese44:
a more robust solution could be to queue up subscriptions and subscribe them once document.body is ready... but i don't think that work is really necessary. if embed needs to start using KeyboardNavigationObserver then we could move its script from head to bottom of body

richardtagger:
Yeah we can probably just move this block https://github.int.yammer.com/yammer/workfeed/blob/master/app/views/layouts/widget.html.erb#L16.

I'm unopinionated on the solution we pick

Issue #243, opened 2018-02-06
Callout is slow


richardtagger:
Looking at flame graphs when debugging typeahead performance reveals hundreds of milliseconds spent recalculating styles in Fabric's Stylesheet.js. The root cause appears to be originating from Callout. Current approach will be to ditch use of Callout for type ahead - we actually don't need the things it provides such as animation, and positioning logic and are better off using an absolutely positioned `DIV with a border radius + shadow. However, future features that are highly interactive and re-render on keypresses may encounter similar issues.

[Chrome Perf Profile.zip](https://github.com/Microsoft/YamUI/files/1700632/Chrome.Perf.Profile.zip)

![screen shot 2018-02-06 at 12 01 18 pm](https://user-images.githubusercontent.com/692240/35881526-eddad898-0b35-11e8-96b1-62e8f50318cc.png)

swese44:
please report this in [Fabric](https://github.com/OfficeDev/office-ui-fabric-react). we get things like positioning + beak for free and it would be ideal to get perf issues fixed in the Fabric repo rather than own this ourselves

richardtagger:
Well we don't use beak or positioning for the editor, but yes we'll use Callout for other things. I will report it to the team

PR #244, opened 2018-02-06
perf(SuggestionsList) Remove Callout, support right-alignment for RTL use-case

richardtagger:
See https://github.com/Microsoft/YamUI/issues/243

We don't use Callout's features such as beak, dynamic positioning or animation. Coupled with the poor performance issues, I've removed the use of it and we can revisit if the performance improves or we need to use more features from Callout


richardtagger:
@unindented @swese44 @adrianmg

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -5,16 +5,18 @@ import { NestableBaseComponentProps } from '../../util/BaseComponent/props';
...
 import './SuggestionsList.css';
-import { IPoint } from 'office-ui-fabric-react/lib/utilities/positioning';
 
 export interface SuggestionsListProps extends NestableBaseComponentProps {
   /**
-   * The target that the Callout should try to position itself based on.
+   * The left offset position in pixels
    */
-  target: HTMLElement | IPoint;
+  left: number;
```
unindented: This is not RTL-friendly.

richardtagger: you're right :(. It's no less RTL friendly than our current stuff but I'll have a think about how we can solve it

unindented:
The hardcoded `left` is a blocker. Other than that it looks good to me.

richardtagger:
Here's a thing https://github.com/Microsoft/YamUI/pull/244/commits/2d35ea9c90644169e5f55bb5c9bf30d321b12b30

richardtagger:
@adrianmg @mamuso the screenshots changed too so I'll need your LGTM please

unindented:
@richardtagger do we need the absolute positioning to live inside this component? Couldn't the consumer position it however it wants?

richardtagger:
gonna close this PR and put together something that is in that direction

PR #245, opened 2018-02-07
feat(SuggestionsList): Remove container

richardtagger:
Removes the containing Callout component because of performance concerns but also to enable flexibility in how it is used. SuggestionsList can now be used in a Fabric surface of choice.

## Pull request checklist


richardtagger:
@unindented @adrianmg

adrianmg:
Rich, I am going to postpone this one until tomorrow (first thing in the morning) since the changed files are so different due to the nature of the change that I'll need to compare and check things by getting the branch and running it locally.

adrianmg:
I'm more than OK with the changes in the implementation but then we need a way to be able to judge and review this component with a real case scenario. Do we want to include a case using the container as part of the component or moving that example to the demo section?

Either way, we need to see the component being used in real life so we can tackle issues that are going to be there like rounded corners of the component colliding with the rounded corners of the container, how wide or not is going to be the final application… etc.

Am I explaining myself?

adrianmg:
@richardtagger this is almost there, I found out a couple of minor changes needed that didn't appear in the last review (probably my fault). Let me know if everything is clear!

image


The rest looks perfect

richardtagger:
Updated

adrianmg:
@richardtagger @unindented I pushed some small changes fixing the 'M Spinner' and also corrected the SuggestionsList issues we mentioned and discussed this morning. Please take a look.

adrianmg:
Yes I'm, Rich! Do I need to approve the review even if I'm the one commiting some of those changes?

richardtagger:
Yes I need your approval as I created the PR

PR #247, opened 2018-02-12
Update dependencies

unindented:
- Bumped version of `react-styleguidist` and `react-styleguidist-visual`.
- Removed stale screenshots, as `react-styleguidist-visual` now errors if there are any.


config/styleguide-visual/config.js
```diff
@@ -15,6 +15,6 @@ module.exports = {
     args: ['--no-sandbox', '--disable-setuid-sandbox'],
   },
   navigationOptions: {
-    waitUntil: 'networkidle',
+    waitUntil: 'networkidle0',
```
unindented: `puppetteer v1.0.0` renamed this.

swese44:
why did visual diffs change height? also i think the newer version of styleguidist fixed some routing issues, are we able to speed up our diffs by not reloading each page?

unindented:
@swese44 I think font rendering has changed with the latest version of Chrome, at least on Linux. You can check with the swipe tool.

And yes, I thought they had fixed routing, but it's broken again. Try going to https://react-styleguidist.js.org/examples/basic/, clicking on an example and going back, and some examples will disappear.

adrianmg:
Everything looks correct, the new rendering is slightly more accurate and similar to how Sketch is rendering Segoe. I wonder if this is coming from a change in the rendering engine or from actually an update of the font-family we're using. Some of the changes are not just small rendering differences but actually tweaking descendants and other things.

Everything looks correct since it's either expanding up or down but keeping all the baselines in the same place.

PR #251, opened 2018-02-21
fix(AsideSection) Remove uppercase and link wrapping

allobast:
- Add new XXSMALL gutter size - 0.2rem
- Add XXSMALL padding examples to Block.md 
- Apply new XXSMALL bottom padding to AsideSection to reduce space between header and border
- Change color of AsideSection border
- Change color of AsideSection header
- Change styling of AsideSection header and link in order to handle ling link test properly (see screenshots)
- Update snapshots of AsideSection tests accordingly

## Pull request checklist

screen shot 2018-02-21 at 10 47 11 am
screen shot 2018-02-21 at 10 46 57 am
screen shot 2018-02-21 at 10 46 32 am

### DIFF
--- a/src/components/AsideSection/AsideSection.css
+++ b/src/components/AsideSection/AsideSection.css
@@ -3,6 +3,12 @@
 @import "src/css/variables/gutters.css";
 
 .y-aside-section--header {
-  border-bottom: 1px solid $borderColor;
-  padding-bottom: $gutterWidth__xSmall;
+  border-bottom: 1px solid $borderColor__secondary;
+  padding-bottom: $gutterWidth__xxSmall;
+  white-space: pre;
+}
+
+.y-aside-section--action {
+  white-space: initial;
+  text-align: right;
 }
--- a/src/components/AsideSection/AsideSection.tsx
+++ b/src/components/AsideSection/AsideSection.tsx
@@ -5,7 +5,7 @@ import { NestableBaseComponentProps } from '../../util/BaseComponent/props';
 import Block from '../Block';
 import { FixedGridColumn, FixedGridRow, GutterSize } from '../FixedGrid';
 import Heading from '../Heading';
-import Text, { TextSize } from '../Text';
+import Text, { TextColor, TextSize } from '../Text';
 import './AsideSection.css';
 
 export interface AsideSectionProps extends NestableBaseComponentProps {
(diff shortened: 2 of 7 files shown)


src/components/AsideSection/AsideSection.css
```diff
@@ -3,6 +3,12 @@
 @import "src/css/variables/gutters.css";
 
 .y-aside-section--header {
-  border-bottom: 1px solid $borderColor;
-  padding-bottom: $gutterWidth__xSmall;
+  border-bottom: 1px solid $borderColor__secondary;
+  padding-bottom: $gutterWidth__xxSmall;
+  white-space: pre;
```
swese44: what happens when there is a very long title, does it just overflow to the right?

can you explain the problem we're trying to solve? if we just want to keep the `action` text from wrapping then we should probably add a "nowrap" option to FixedGridColumn, we've needed that in other places too

src/components/AsideSection/AsideSection.tsx
```diff
@@ -27,11 +27,11 @@ export default class AsideSection extends React.Component<AsideSectionProps, {}>
   render() {
     return (
       <div className="y-aside-section">
-        <Block textSize={TextSize.SMALL}>
+        <Block>
           <FixedGridRow className="y-aside-section--header" bottomSpacing={GutterSize.SMALL}>
-            <FixedGridColumn>
+            <FixedGridColumn fixed={true}>
```
swese44: why are we moving the smaller column to no longer be fixed?

src/components/AsideSection/AsideSection.tsx
```diff
@@ -27,11 +27,11 @@ export default class AsideSection extends React.Component<AsideSectionProps, {}>
   render() {
     return (
       <div className="y-aside-section">
-        <Block textSize={TextSize.SMALL}>
+        <Block>
```
swese44: you can give Block `push={2}` instead of adding the 2px of CSS

src/components/Block/Block.css
```diff
@@ -27,6 +27,9 @@
 .y-block__topSpacing-xSmall {
   margin-top: $gutterWidth__xSmall;
 }
+.y-block__topSpacing-xxSmall {
```
swese44: we should remove all the xxSmall changes, 2px doesn't map to our 4px GutterSize (which is going to be replaced by a scalar number in the future)

src/components/AsideSection/AsideSection.css
allobast: I tried adding white-space: nowrap; to all FixedGridColumn, it doesn't wrap but looks like: 
screen shot 2018-02-21 at 6 13 09 pm

My changes allow text to grow until it meets another column, and only then jump to the next line
screen shot 2018-02-21 at 6 14 12 pm

src/components/AsideSection/AsideSection.tsx
allobast: With a fixed={true} fake link is aligned to right, even though I set text-align: right;
I just played with it for a while, tested different scenarios and came to this combination of CSS and components properties

src/components/AsideSection/AsideSection.tsx
allobast: Well, it won't really do what we want, we need spacing between header and border. Applying push={2} to the above Block doesn't do the thing

src/components/AsideSection/AsideSection.css
swese44: cool thanks. let's ask Design how this should behave

allobast: I confirmed all the behavior with @adrianmg and showed him what it looks like now after changes :)

chrismohr:
Checked with Alina, closing due to staleness.

PR #253, opened 2018-02-22
Bump typescript version to 2.7.2

alxHenry:
Bump typescript version to latest. Doing this to fix a typescript issue where you can't set an enum value as the value of another enum. We would like to use that to limit the number of fabric enum options they have.

## Pull request checklist


src/util/DebouncedOnChange/index.tsx
```diff
@@ -60,7 +60,7 @@ export default class DebouncedOnChangeComponent extends React.Component<
       this.props.onChange(newValue);
     }
 
-    if (this.props.debouncedOnChange) {
+    if (this.props.debouncedOnChange && this.debouncedOnChange) {
```
unindented: Why is this one necessary if you're initializing it in the constructor?

alxHenry: It's only initialized if props.debouncedOnChange is true. Typescript doesn't realize that props.debouncedOnChange being true is equivalent to this.debounceOnChanged being defined, so it complains that it may be undefined.

PR #254, opened 2018-02-23
Add MenuButton

alxHenry:
- Add the MenuButton component
- Add the MenuButtonIcon component that we pass to fabric as a custom renderer for menu items
- For now the only type of target will be an icon. In the future there will be a followup to add a button as a target for the menu.
- Fix a small existing issue that styleguidist was throwing console errors for
- Expose some additional existing text colors

## Pull request checklist


alxHenry:
@swese44 @unindented we can start taking a look at the code. Once the fabric release is cut (should be tomorrow) I can do the visual diff stuff.

alxHenry:
@mrfunkycold

alxHenry:
@unindented @swese44 I bumped Fabric to the version with my changes and everything still looks good here and in Dropdown.

Added the initial visual diffs as well. Unfortunately fabric doesn't have a 'render with menu open' or anything. @unindented what is the repo for the screenshotter you made? Might take a look at the robo click of elements.

unindented:
@alxHenry https://github.com/unindented/react-styleguidist-visual

alxHenry:
@swese44 updated the tests, completed coverage, added the coloring of the icons

src/components/MenuButton/MenuButton.css
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/* Element styling overrides */
+.y-menu-button--callout .ms-ContextualMenu-link,
+.y-menu-button--callout .ms-ContextualMenu-header {
+  padding: 0;
+}
+
+.y-menu-button--item {
+  padding: .4rem 1.2rem .8rem
```
swese44: semicolon at end of line

src/components/MenuButton/MenuButton.css
```diff
@@ -0,0 +1,58 @@
...
+}
+.y-menu-button--item-header {
+  padding: .6rem 1.2rem 1rem
```
swese44: semicolon at end of line

src/components/MenuButton/MenuButton.css
```diff
@@ -0,0 +1,58 @@
...
+  background-color: initial;
+}
+
+.y-menu-button--callout .ms-ContextualMenu-divider {
+  background-color: $borderColor__secondary;
+  margin-bottom: -1px;
+}
+
+
+.y-menu-button--callout a:hover {
```
swese44: we should replace these `a` tag selectors with the classname of the `a` and `button` tags. also want to make sure the bgcolors are correctly the same on both tags

src/components/MenuButton/MenuButton.md
```diff
@@ -0,0 +1,76 @@
...
+      icon: Reply,
+      isDisabled: true,
+      onClick: () => console.log('icon and disabled'),
+    }, {
+      key: 'header',
+      text: 'Header Type with Divider After',
+      type: 2,
+    }, {
+      key: 'divider',
+      type: 1,
```
swese44: are we unable to use the enum here in the .md file instead of the compiled int value?

src/components/MenuButton/MenuButton.md
```diff
@@ -0,0 +1,76 @@
...
+  <MenuButton
+    ariaLabel="aria"
+    menuItems={[{
+      key: 'regular',
+      icon: Edit,
+      text: 'Icon Item',
+    }, {
+      key: 'noIcon',
+      text: 'No Icon',
+      onClick: () => console.log('no icon'),
```
swese44: we should replace these `console.log()` statements with `action()`, other component example files should have this

src/components/MenuButton/MenuButton.md
```diff
@@ -0,0 +1,76 @@
...
+      href: 'http://bing.com',
+      onClick: () => console.log('link'),
+    }]}
+  />
+</div>
+```
+
+MenuButton with specified Icon and basic children
+
+```js { "props": { "data-description": "specified icon and icon size" } }
```
swese44: let's fix the description, maybe just "custom icon"?

src/components/MenuButton/MenuButton.md
```diff
@@ -0,0 +1,76 @@
+### Notes for use
+
+Context menu should be used as a menu button where you want to provide options that are hidden behind a click.
+
+### Examples
+
+MenuButton with default Icon, and all applicable children properties
+
+```js { "props": { "data-description": "with deafult icon and children properties" } }
```
swese44: "default" is misspelled.

how about "with default icon and all menu item types"

src/components/MenuButton/MenuButton.md
swese44: you'll probably have to delete these screenshots from `config/styleguide-visual/reference` fyi

src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  isDisabled?: boolean;
+  onClick?: (ev?: React.MouseEvent<HTMLElement>) => void;
+  href?: string;
+  icon?: typeof BaseIcon;
+}
+
+export interface MenuButtonProps extends BaseComponentProps {
+  ariaLabel: string;
+  menuItems: MenuButtonItem[];
+  className?: string;
```
swese44: we can delete className, it gets inherited from BaseComponentProps

src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+import './MenuButton.css';
+
+export { MenuItemType } from './enums';
+
+const MoreIcon = require('../Icon/icons/More').default;
+
+const renderEmptyIcon = () => null;
+
+export interface MenuButtonItem {
+  key: string;
```
swese44: We should add comments before each item in the `MenuButtonItem` and `MenuButtonProps` interfaces, this will show up in the editor as TypeScript intellisense descriptions

src/components/MenuButton/MenuButton.css
alxHenry: These a tags have the same class as all the items, .ms-ContextualMenu-link. I replaced the a selector with that class and I think it should be okay.

alxHenry: bgcolors look the same for link and non link

adrianmg:
I've made some small adjustments here: https://github.com/Microsoft/YamUI/pull/254/commits/b569740e14b94832b847eacad85d64bee8226318

Sharing some questions:
- Could we generate more examples with different combinations with the titles and no titles? So I can check all the different options and margins.
- Is someone taking care of the document list case?

Pending things to be fixed after my commit:
- Row with icon: Move icon 1px down VS label
- Border radius is not applied to the general container (or not visible)
- The disabled label looks bolded instead of regular (this thicker even if the opacity is lower)
- Header shouldn't have 2 line dividers (check design spec)
- Why are we hiding the component whenever I press Alt/Cmd? I'm wondering if that's an accessibility issue since I may want to Cmd+Click to open a section in a new tab and that's preventing me from doing that. @BrendanMcK 
- The default is not the one we defined, I'm actually thinking on following this (re-open and check Zeplin small update on this):
- 1. Default state: defined by CSS
- 2. Hover: defined by CSS
- 3. Active: keep the hover state and force 100% opacity (unless something else custom CSS is defined ad-hoc)

The rest looks OK 👍

adrianmg:
@alxHenry clarifying some of those questions!

- Generating more cases (image attached): Having the header in different places so make sure the margins are always kept and flowing right with the vertical rhythm could be helpful.
- +1 creating an issue ticket for us/Fabric to fix the concerns regarding accessibility.
- I'm happy with defining '•••' as the default one.
- Regarding states: currently when the icon is active it loses opacity, my idea is to keep the same styling as the hover state and make sure opacity is 100% as the default behavior. All the three states (default, hover and active) should be easy to override if needed at some point in CSS.

image

alxHenry:
@adrianmg are you referring to the icon being blue when the element is disabled?

I'm not seeing the icon ever have less than 100% opacity with the active state. Am I reading that right?

adrianmg:
@alxHenry sorry! Yes, I thought we were reducing the opacity because it seemed a lighter blue compared to the hover state, sorry!

alxHenry:
No worries, will fix that color

alxHenry:
@adrianmg I changed the examples a bit to have headers and dividers and both to test out cases.  Also updated the item icon to be gray when disabled.

I think that's everything, let me know if it isn't.

src/components/MenuButton/MenuButtonItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return join(classes);
+};
+
+const MenuButtonItem: React.StatelessComponent<IContextualMenuItemProps> = (
+  props: IContextualMenuItemProps,
+) => {
+  const { data, name, itemType, disabled } = props.item;
+
+  const Icon = data && data.yamUIIcon;
+  const iconColumn = Icon ? (
```
swese44: it might be a little cleaner to move this to a private `getIconColumn()` method

src/components/MenuButton/MenuButton.css
```diff
@@ -0,0 +1,80 @@
...
+  color: $textColor__secondary;
+}
+.y-menu-button:hover,
+.y-menu-button:active,
+.y-menu-button:focus {
+  color: $linkColor;
+  opacity: 1;
+}
+
+.y-menu-button--callout {
```
unindented: I think this is the opposite of BEM. If we were following BEM, this would be `y-menu-button__callout`.

unindented: http://getbem.com/naming/

adrianmg:
@alxHenry disabled rows shouldn't be interactive (it currently shows an hover state).

PR #255, opened 2018-02-23
Use aria-required attribute for Textfield instead of required

chrismohr:
`required` was triggering undesired browser behaviour. `aria-required` will still announce that the field is required without the undesired side effects. 

- added benefit is that I was also able to remove a style override that was hiding the asterisk that came along with `required`

yammer___screen_shot_2018-02-16_at_3_13_03_pm

### DIFF
--- a/src/components/BaseTextField/BaseTextField.css
+++ b/src/components/BaseTextField/BaseTextField.css
@@ -38,9 +38,3 @@
   color: $textColor__error;
   padding-top: .2rem;
 }
-
-/* Hide the asterisk that comes with required=true */
-.y-base-text-field .ms-Label::after,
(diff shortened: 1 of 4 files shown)


swese44:
should we update the stories to use `aria-required` instead of `required`? also does this affect TextArea?

chrismohr:
@swese44 what part of the stories did you have in mind?  I was thinking nothing needs to change, since the external interface still calls the prop `required` and the description still seems good: `Use to indicate that a value must be provided to allow the containing form to be submitted.`

There's an example for both a required `TextField` and `TextArea`, and the snapshots have updated to reflect the new implementation.

PR #256, opened 2018-02-26
Adding all the document icons

mamuso:
Adding ODSP document icons


unindented:
@mamuso Are these optimized in any way? I'm seeing Illustrator comments in there, which probably means they haven't gone through SVGO.

unindented:
@mamuso ah, nevermind. It's our tasks that are the problem, we're only reading from the `icons` folder.

mamuso:
@jessemartin is going to need this for his project. @unindented, @swese44, @richardtagger what would be the best way to move this component forward?

unindented:
@mamuso this PR has everything green, so feel free to press the button.

If Townhall needs the component, @jessemartin can probably pair with P.J. or I to implement it.

PR #257, opened 2018-02-27
Add tabindex=-1 and aria-hidden="true" to icons

BrendanMcK:

## Pull request checklist


config/icons/iconTemplate.ejs
```diff
@@ -6,11 +6,9 @@ import * as React from 'react';
 import BaseIcon from '../BaseIcon';
 
 export default class ${name} extends BaseIcon {
-  render() {
+  renderPath() {
     return (
-      <svg viewBox="0 0 240 240" className={this.getClassName()} style={this.getStyle()}>
-        ${jsx}
-      </svg>
+      ${jsx}
```
unindented: Could create invalid React 15 components if an icon had more than one element inside its `svg` tag...

BrendanMcK: Good point about potential invalid items. Given that multiple paths are perfectly legit, I'll revert to the more verbose version; can revisit and use fragments when we're using React 16. Otherwise it's an obscure build error waiting to happen should someone add an icon with two paths in future.

BrendanMcK:
Updated version force-pushed. LMK know if this still looks good.

Issue #167, opened 2017-12-16
SuggestionsList missing template for results other than User


swese44:
type PickerItemTemplate = 'text' | 'mediaRound' | 'mediaSoft' | 'mediaSquare';

mediaRound: Existing Avatar implementation
mediaSoft: Add border="soft" to Avatar
mediaSquare: Use Image instead of Avatar
text: Plain text aligned left

PR #260, opened 2018-02-28
fix(Avatar,SuggestionList): Fix Avatar badging, add avatarProps to SuggestionList

richardtagger:
Updates Avatar badging CSS, adds prop to SuggestionsListItem props

## Pull request checklist


richardtagger:
@swese44 @adrianmg ready to review I think

src/components/SuggestionsList/SuggestionsListItem.tsx
```diff
@@ -5,13 +5,17 @@ import { NestableBaseComponentProps } from '../../util/BaseComponent/props';
...
+import { IconSize } from '../Icon';
+
 import './SuggestionsListItem.css';
 
 export interface SuggestionItem {
   id: string | number;
   imageUrl?: string;
   name: string;
   description?: string;
+  isFlexternal: boolean;
```
swese44: i really don't like that this would bake Yammer business logic directly into SuggestionsList, YamUI should be enforcing the visuals/layout but not the actual content like "BadgeFlexternal".

We're also not yet looking at this holistically because we haven't implemented the other SuggestionsListItem templates `text` | `mediaRound` | `mediaSoft` | `mediaSquare`: https://github.com/Microsoft/YamUI/issues/167

i think i would prefer SuggestionsListItem to accept badgeContent or AvatarProps, somehow exposing a way for the client to set badgeContent without dictating what that content should actually be. it seems that only `mediaRound` will optionally allow badge content via Avatar, so this specific item interface could also accept AvatarProps?

swese44: `isFlexternal` is a Yammer business concern, not a presentational concern

swese44: additionally we'd have to keep on adding new props for every new badge type we want to manually support here...

src/components/SuggestionsList/SuggestionsListItem.css
```diff
@@ -6,6 +6,13 @@
   padding: 0.4rem 1.2rem 0.8rem 1.2rem;
 }
 
+.y-suggestion-list-item--badge {
+  background-color: $badgeColor__background;
+  fill: $badgeColor__fill;
+  display: block;
```
swese44: set block={true} on Icon

src/components/SuggestionsList/SuggestionsListItem.css
```diff
@@ -6,6 +6,13 @@
   padding: 0.4rem 1.2rem 0.8rem 1.2rem;
 }
 
+.y-suggestion-list-item--badge {
+  background-color: $badgeColor__background;
```
swese44: can delete this line after @adrianmg's change

src/components/SuggestionsList/SuggestionsListItem.css
```diff
@@ -6,6 +6,13 @@
...
+  fill: $badgeColor__fill;
+  display: block;
+  top: 0;
```
swese44: if YamUI Icon always needs to be adjusted here, we should add a rule in Avatar to handle it instead:

```
.y-avatar--badgeContent > .y-icon {
  top: 0;
}
```

not ideal, but better than a one-off for SuggestionsList's Flexternal case

src/components/SuggestionsList/SuggestionsListItem.tsx
richardtagger: hmm - calculating isFlexternal is Yammer business logic, but supporting a badge type called Flexternal isnt particularly. It's just a dev convenience. Could change this type a badgeType?: 'globe' | 'star' | 'face' if that helps?

src/components/SuggestionsList/SuggestionsListItem.css
```diff
@@ -6,6 +6,13 @@
   padding: 0.4rem 1.2rem 0.8rem 1.2rem;
 }
 
+.y-suggestion-list-item--badge {
+  background-color: $badgeColor__background;
+  fill: $badgeColor__fill;
```
swese44: i don't think we should be enforcing this here, instead let the client set badgeContent in avatarProps for the item template

src/components/SuggestionsList/SuggestionsListItem.tsx
swese44: badge content is not the concern of SuggestionsList, it's really the concern of Avatar. even adding badgeType prevents SuggestionsList from being used in different ways. what if we use SuggestionsList to search for attachments, and we want badgeContent to be the file type icon? what if we want to show people's presence as the badge content? i don't think SuggestionsList should be gatekeeper, it's displaying an Avatar so i think Avatar should be the content gatekeeper.

i totally agree that "Flexternal" is a bad name for an icon, maybe we should separately rename that to something generic like "globe"

richardtagger: I'm fine with us avoiding Yammer domain terms like Flexternal. It seems like `SuggestionItem` should extend `AvatarProps`. Agree?

swese44: i think it would be better for SuggestionItem to accept an avatarProps field, and SuggestionItem could splat them into its Avatar before setting its own props over the top

richardtagger: ok

richardtagger:
@swese44 how about now?

richardtagger:
also @adrianmg

adrianmg:
Sorry, I just detected a few minor things related to the size of the SVG. Let me dig a bit further into this and I'll let you know if changes are needed

adrianmg:
@swese44 @richardtagger I see what's happening… I was comparing the Avatar with Badge VS the Avatar with Badge within the SuggestionsList.

Should we update the Avatar component and use the real SVG badge now we're implementing it properly (and not faking it with a plain PNG)?

src/components/Avatar/Avatar.css
```diff
@@ -49,16 +50,22 @@
...
 }
 .y-avatar__size-small--badge,
 .y-avatar__size-xSmall--badge {
   height: 16px;
   width: 16px;
 }
 
+.y-avatar--badge > .y-icon {
+  top: 0;
+  display: block;
```
richardtagger: we'll wanna use Icon prop block=true here

adrianmg: Does it keep the fix visually? I'm not familiar with what that does

richardtagger: thats fine. I'll make the change and the visuals shouldnt change.

richardtagger:
Done - Block'd the example where we pass badgeContent and I've deleted badge.png

Issue #261, opened 2018-02-28
Text truncation doesn't behave consistently across browsers


richardtagger:
Issue is how IE/EDGE handles truncation of inline-block items - it seems to treat them as a ~single character~, so truncates the entire inline-block down to an ellipsis. If display: inline is used here, all's well.  Firefox does similar - seems that Chrome is the outlier here.

More context:-
1. we use display: inline-block on <Text> because the browser will ignore line-height on inline elements if the nearest block-level ancestor has a larger line-height.
2. we also need display: inline-block on <Text> when we ellipse text with "maxWidth", it doesn't work on an inline element.
3. additionally, we use vertical-align: baseline to line up the baseline of text characters regardless of font-size, line-height, etc. BUT the browser says "nope" and forces vertical-align: bottom when overflow:hidden is added for maxWidth-ellipsed text. this throws off our vertical rhythm between different text on the same line, so we have rem vertical-align adjustments for these cases.
4. but i've found that these are not reliable across browsers, Safari will adjust adjacent text higher or lower so i'm sure IE/Edge also has some wonkiness.

long story short, we might want to rethink <Text>, possibly doing the following:
1. set <Text> to display: inline
2. remove the maxWidth option from <Text>, find a different way to ellipse complicated groups of inline text (like in our message byline)
3. just have to know that sized <Text> will match the vertical-align of its closest <Block> ancestor with a specified "textSize".

PR #262, opened 2018-02-28
fix(suggestion-list): Replace use of Text with Spans

richardtagger:
Works around an issue with ellipses truncation in browsers other than Chrome. See Issue #261 

## Pull request checklist


src/components/SuggestionsList/SuggestionsListItem.tsx
```diff
@@ -22,16 +21,21 @@ export interface SuggestionsListItemProps extends SuggestionItem, NestableBaseCo
...
 const baseClass = 'y-suggestions-list-item';
 const selectedClass = `${baseClass} y-suggestions-list-item--selected`;
+const matchHighlightClass = 'y-suggestions-list-item--match-highlight';
 
 const getHighlightedName = (name: string, search: string) => {
-  return name.split(new RegExp(`(${search})`, 'gi')).map((item: string, index: number) => {
-    const isBold = search.toLowerCase() === item.toLowerCase();
-    return (
-      <Text key={index} bold={isBold}>
-        {item.replace(/\s/g, '\u00A0')}
```
swese44: are we sure this problem wasn't caused by replacing spaces with non-breaking-space characters, rather than the Text component?

richardtagger: checking

richardtagger: Here's what that was addressing for `Text`
screen shot 2018-02-28 at 3 32 00 pm

richardtagger:
@adrianmg

src/components/SuggestionsList/SuggestionsListItem.tsx
swese44: ah :+1:

adrianmg:
@richardtagger I wonder why the initials avatar is just showing up one letter?

image

richardtagger:
Because the person has like 5 names - I guess that's how Fabric's Persona handles that

PR #269, opened 2018-03-02
Release version 5.1.1

alxHenry:
- Add some icons
- Add the new MenuButton component

- I'm dumb and copy pasted patch instead of minor from the readme

## Pull request checklist


alxHenry:
@swese44 here's the version bump. Technically should be a minor instead of a patch. We can fix if you want

PR #271, opened 2018-03-02
Update README.md to make version bumping more clear

BrendanMcK:

## Pull request checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,13 +86,17 @@ It's best to have at least 30GB of free space for Docker containers and images. 
 
 ## Releasing a new version
 
-If you want to release a new version of YamUI, do the following to create a release branch:-
+If you want to release a new version of YamUI, do the following to create a release branch. Replace version_type with `major`, `minor`
+or `patch` as appropriate, based on [SemVer](https://semver.org):-
 
 ```sh
 git checkout -b awesome-release-branch
-npm version patch -m "Release version %s"
+npm version version_type -m "Release version %s"
 git push
+git push --tags
 ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,14 +86,18 @@ It's best to have at least 30GB of free space for Docker containers and images.
...
 
-If you want to release a new version of YamUI, do the following to create a release branch:-
+If you want to release a new version of YamUI, do the following to create a release branch. Replace version_type with `major`, `minor`
+or `patch` as appropriate, based on [SemVer](https://semver.org):-
+
 
 ```sh
 git checkout -b awesome-release-branch
-npm version patch -m "Release version %s"
+npm version version_type -m "Release version %s"
```
swese44: maybe wrap use backticks around \`version_type\` for emphasis?

alxHenry:
I approved, but I don't have any power here for ultimate approval :(

README.md
```diff
@@ -86,14 +86,18 @@ It's best to have at least 30GB of free space for Docker containers and images.
...
 git push
```
swese44: should probably also add `git push --tags` after `git push`

README.md
BrendanMcK: I tried that - but neither that nor underscores nor HTML works within a code block :-/ So I went with something that would fail if copy/pasted.

README.md
BrendanMcK: Ah, interesting. Is that needed instead or in addition to the git push?

BrendanMcK: From reading git manual stuffs, seems its needed as a separate step. Updated.

PR #272, opened 2018-03-03
ContainerQuery component for responsive features

### LINKED ISSUE #216: Provide a container query solution for responsive components
[react-container-query](https://github.com/d6u/react-container-query) could be the best solution, but it weighs in at 42kb. we could contribute to lighten this footprint by replacing lodash.isequal with something tiny to compare objects, and possibly update the resize-observer-lite dependency:
https://github.com/d6u/react-container-query/issues/66

swese44:
Adds `ContainerQuery`, a component to allow conditional rendering based on the component's height/width. Basically a media query solution but at the element level instead of the full page.

containerquery

`react-container-query` adds 22kb of minified JavaScript so we may want to be selective about where we use this.

fixes #216

## Pull request checklist


unindented:
Is `react-container-query` doing something stupid like including a polyfill? Is it 24kb uglified, or 24kb gzipped? It's a lot in any case.

unindented:
Yup, confirmed. It's including `resize-observer-lite`, which includes `element-resize-detector`, which is an 18kb polyfill. Opened an issue: https://github.com/d6u/react-container-query/issues/71

swese44:
looks like the `react-container-query` dist is 22kb minified/uglified (i saw a 24kb increase in overall bundle size).

FYI only Chrome supports ResizeObserver, it's being iterated on in WICG.
https://caniuse.com/#feat=resizeobserver
https://wicg.github.io/ResizeObserver/

unindented:
Right, but by bundling the polyfill with the library they're forcing us to parse and execute 18kb of unnecessary code on Chrome (and soon Firefox and Edge).

adrianmg:
Is this ready for review, guys? I'm seeing this is for other purposes but just wanted to check if you need a sign-off from me

swese44:
@adrianmg we do need Design signoff, but we're also going to hold off until we can decrease the asset size of the external library this would introduce.

swese44:
i've proposed a solution for the `element-resize-detector` dependency:
https://github.com/d6u/react-container-query/issues/71

unindented:
We ended up implementing this in `SizeObserver`. I'm gonna close.

PR #273, opened 2018-03-06
Additional MenuButton Options

alxHenry:
- Add an iconSize option to choose the size of the click target icon
- Add target option for a menu item. Sets the target attribute of the anchor tag that is produced. Some places in Yammer open in the same tab, others open a new tab.
- Get test coverage to 100%, I had missed the onRenderMenuIcon method that simply returns null

screen shot 2018-03-06 at 9 59 47 am

## Pull request checklist


alxHenry:
@swese44 @unindented

unindented:
- Can you paste a manual screenshot of the thing showing up?
- Multiple icon sizes is a design requirement?
- Do we have to allow a specific target? Can't we reduce the posibilities to a boolean (new window or current window) like we do in the current link component?

alxHenry:
- The only visual change in this PR is for the size of the target icon, attached a screenshot comparing
- The design spec did not have an icon size prop. The default target in the spec is 20 (Large), places in Yammer are kind of random, I've noticed 12 (Xsmall) and 14 (Small). It seemed like it wouldn't be long before we wanted an option, but I'm also fine with making the default target icon size 14px.
- The boolean seems fine to me

src/components/MenuButton/MenuButton.tsx
```diff
@@ -109,6 +119,7 @@ export default class MenuButton extends React.Component<MenuButtonProps, {}> {
       onClick: item.onClick,
       disabled: item.isDisabled,
       href: item.href,
+      target: item.newWindow ? '_blank' : undefined,
```
swese44: can we also pass `rel="nofollow noopener noreferrer "` here to prevent tab hijacking attacks?
https://github.com/Microsoft/YamUI/blob/master/src/components/NavigationLink/NavigationLink.tsx#L37

src/components/MenuButton/MenuButton.tsx
```diff
@@ -77,6 +82,11 @@ export interface MenuButtonProps extends BaseComponentProps {
    * YamUI Icon to display as the click target. Default is the 'More' icon.
    */
   icon?: typeof BaseIcon;
+
+  /**
+   * Size of the icon. Defaults to large.
```
swese44: we should probably use the @default syntax for default values

src/components/MenuButton/MenuButton.tsx
```diff
@@ -94,10 +104,10 @@ export default class MenuButton extends React.Component<MenuButtonProps, {}> {
 
   @autobind
   private getIcon() {
-    const { icon } = this.props;
+    const { icon, iconSize = IconSize.LARGE } = this.props;
     const Icon = icon || MoreIcon;
 
-    return <Icon size={IconSize.LARGE} block={true} />;
+    return <Icon size={iconSize} block={true} />;
```
swese44: isn't there CSS on the IconButton setting it to 20px by 20px? do we need to map the sizes in CSS to the icon size?

src/components/MenuButton/MenuButton.tsx
alxHenry: Doesn't look like Fabric supports this, we'd need to add the prop in their repo. They create the anchor tag for us

src/components/MenuButton/MenuButton.tsx
alxHenry: Yeah we did. Will have to map each icon size enum member to a css class I guess?

src/components/MenuButton/MenuButton.tsx
alxHenry: https://github.com/OfficeDev/office-ui-fabric-react/issues/4201

alxHenry:
Merged the fabric PR to fix security concern when passing target '_blank' today. Will bump Fabric when it's released and then merge this PR in.

swese44:
nice!

Issue #195, opened 2018-01-08
Visual diffs depend on CDN Segoe font, which is not always reachable


jessemartin:
Last week I ran the visual snapshot tool on master via`y run test:visual` and most of the tests failed for font-related reasons. Further investigation with @unindented revealed that the Segoe font was not loading properly.

We should save a copy of the Segoe font in YamUI and reference it locally to prevent network flakiness from failing our test suite.

jessemartin:
@swese44 @unindented

unindented:
Agree.

PR #280, opened 2018-03-10
SVG Illustrations component and build task WIP

swese44:
- [x] Get case-sensitive icon filenames from Design
- [x] Get updated illustration svg files without `<style>` tags from Design
- [x] Merge icon and illustration svg tasks, reduce duplication

## Pull request checklist


config/illustrations/illustrationTemplate.ejs
```diff
@@ -0,0 +1,24 @@
...
+  render() {
+    return (
+      <svg
+        viewBox="0 0 ${size} ${size}"
+        height="${size}"
+        width="${size}"
+        className={this.getClassName()}
+        aria-hidden="true"
+        tabIndex={-1}
+        focusable="false"
```
BrendanMcK: Nice. I'm happy.

jessemartin:
@swese44 How's this look to you?

@mamuso I think we just need the updated icons (without style tags and with updated file names)

config/svg/generateIllustrations.js
```diff
@@ -0,0 +1,55 @@
...
+
+const optimizer = new SVGO(svgoConfig);
+
+function getSizeFromFilePath(filePath) {
+  const sizeRegex = /.+\/(\d+)\/.+\.svg/;
+  const sizeResults = filePath.match(sizeRegex);
+
+  if (sizeResults === null) {
+    console.error(`Could not determine svg size from path: ${filePath}`);
+  }
```
unindented: Should you be returning or throwing here? Otherwise you're going to get a null reference exception in line 30.

package.json
```diff
@@ -80,6 +80,8 @@
     "clean:dist": "rimraf dist/*",
     "clean:docs": "rimraf build/docs/*",
     "clean:icons": "rimraf build/icons/* src/components/Icon/icons/*.tsx",
+    "clean:illustrations": "rimraf build/illustrations/* src/components/Illustration/icons/*.tsx",
```
unindented: Is this path correct? `src/components/Illustration/icons/*.tsx`

jessemartin: Good catch, I'll update.

config/svg/generateIllustrations.js
jessemartin: In the case that there is an invalid SVG path, I think we'll want to throw and stop the build. @swese44 and I verbally discussed this but forgot to translate it to code. I'll update.

config/styleguide/config.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
     '**/*.spec.tsx',
     '**/components/Fabric/*',
     '**/components/Icon/icons/*',
+    '**/components/Illustration/sizedIllustrations/*',
```
swese44: looks like this path needs to be updated to:
`'**/components/Illustration/illustrations/*',`

config/svg/indexTemplate.ejs
```diff
@@ -0,0 +1,4 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+
+<% svgNames.forEach((svgName) => { %>export { default as <%= svgName %> } from './<%= svgName %>';
+<% }) %>
```
swese44: we should add a newline at the end here

swese44:
looking good! 2 minor things, and we'll just want to verify our outputs once we get the new svg files

config/svg/indexTemplate.ejs
jessemartin: Done

jessemartin:
@swese44 @unindented Any other comments?

src/components/Illustration/BaseIllustration.tsx
```diff
@@ -0,0 +1,36 @@
...
+import * as React from 'react';
+import { BaseComponentProps } from '../../util/BaseComponent/props';
+import { IllustrationSize } from './enums';
+import './Illustration.css';
+
+export { IllustrationSize };
+
+export interface IllustrationProps extends BaseComponentProps {
+  /**
+   * Sets icon style to `display: block`. This is helpful if the inline icon's `line-height` is
```
unindented: Comment here mentions `icon`.

src/components/Illustration/docs/Illustrations.md
```diff
@@ -0,0 +1,40 @@
+### Notes for use
+
+Because current bundlers do not always tree-shake properly, illustrations are exported individually, so you'll need to import each illustration separately.
+
+Many illustrations, like our file type icons, have different
```
unindented: This line seems incomplete.

src/components/Illustration/enums.ts
```diff
@@ -0,0 +1,9 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+export enum IllustrationSize {
+  XXLARGE = '72',
+  XLARGE = '48',
+  LARGE = '40',
+  MEDIUM = '32',
+  SMALL = '20',
+  XSMALL = '16',
+}
```
unindented: What's this used for?

jessemartin: I don't think it is used. @swese44 I'm going to delete it for now. If you had a use in mind, let us know!

jessemartin:
Awesome, thanks @mamuso

jessemartin:
Looks like quite a few visual diffs are failing now. This could be caused by the Segoe font not loading properly.

I'll run it locally and investigate.

jessemartin:
Locally, I'm only getting one failure:
screen shot 2018-03-19 at 12 51 16

jessemartin:
https://github.com/Microsoft/YamUI/issues/195 :(

swese44:
@jessemartin @unindented we could include the font files in our repo and link to them only for visual diff tests (not when running the app locally or in the generated docs for ghpages). the tradeoff would be more reliable visual diff runs, but a potential gap in coverage for the actual font files

swese44:
@jessemartin Alex made a recent change to this file, could be related. feel free to just override the file with your new screenshot, that's only a demo component and its height doesn't matter

unindented:
@swese44 font files probably won't change that often, so maybe it's the right tradeoff?

swese44:
ya i'm fine with it since visual tests have been so flakey. i think we want to only include those local font files when the app is running for visual diffs, i think we have some other conditional CSS in the same place to disable CSS animations

jessemartin:
I've rebased this branch and cleaned up the commits. @swese44 @unindented Mind looking it over again? You should be able to review by commit.

swese44:
woohoo!

PR #284, opened 2018-03-20
fix(SuggestionListItem): Prevent SuggestionListItem from stealing focus

richardtagger:
When clicking SuggestionListItems the mousedown event should be preventDefault’d so the focused item in the page doesn’t get blurred

### DIFF
--- a/src/components/SuggestionsList/SuggestionsListItem.test.tsx
+++ b/src/components/SuggestionsList/SuggestionsListItem.test.tsx
@@ -59,13 +59,20 @@ describe('<SuggestionsListItem />', () => {
   });
 
   describe('when selecting', () => {
+    let preventDefault: jest.Mock;
+
     beforeEach(() => {
       rendered = shallow(<SuggestionsListItem {...getProps()} />);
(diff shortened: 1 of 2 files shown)


richardtagger:
@swese44 @unindented

src/components/SuggestionsList/SuggestionsListItem.test.tsx
```diff
@@ -59,13 +59,20 @@ describe('<SuggestionsListItem />', () => {
   });
 
   describe('when selecting', () => {
+    let preventDefault: jest.Mock<{}>;
```
unindented: `{}` is the default type argument, so no need to provide it.

richardtagger:
done

src/components/SuggestionsList/SuggestionsListItem.test.tsx
richardtagger: done

swese44: my upcoming linting change will address the rest of these...

PR #290, opened 2018-03-27
feat(bistate-button): Add BistateButton component

richardtagger:
Utility wrapper around Button to abstract active / inactive rendering. We might move away from this idiom at some point in the future but until we make that call we'll probably need this

examples:-

screen shot 2018-03-27 at 12 57 42 am
screen shot 2018-03-27 at 12 57 46 am
screen shot 2018-03-27 at 12 57 26 am
screen shot 2018-03-27 at 12 57 33 am

## Pull request checklist


richardtagger:
@unindented @swese44 @adrianmg

richardtagger:
@BrendanMcK this supports blur/focus and mouse enter/leave independently and if you do one whilst the other is active it wont interfere

BrendanMcK:
@richardtagger - nice!

Two things that might be worth seeing if we can address related to this change:

- avoiding the flickering that can happen when the active/inactive states have different lengths.

- I might want to see if we can do better for the accessible name for these. Thinking here is that from screenreader POV, we'd like this to have accessible text of "Joined - Leave" - so that a screenreader user searching for "Leave" can find this, regardless of visual state. I'll do a follow-up PR for this.

richardtagger:
> avoiding the flickering that can happen when the active/inactive states have different lengths.

I've defaulted to fixedWidth=true - that should do it, right?

> we'd like this to have accessible text of "Joined - Leave"

ariaLabel is a customizable prop for active/inactive states, do we need to do more than that?

src/components/Button/BistateButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export type CustomizableProps = 'color' | 'text' | 'ariaLabel' | 'icon' | 'iconPosition' | 'className';
+
+export type CustomizableButtonProps = Pick<ButtonProps, CustomizableProps>;
+
+export interface BistateButtonOwnProps {
+  activeProps: CustomizableButtonProps;
+}
+
+export type BistateButtonProps = BistateButtonOwnProps & ButtonProps;
```
unindented: I lost track of the type manipulation in lines 6-14. Can you explain a bit?

richardtagger: So BistateButtonProps is just ButtonProps + 1 prop `activeProps` which is a subset of ButtonProps; the 'customizable button props' that you might want to vary when the button is 'active'

richardtagger: Can you think of better naming / an easier interface. It's obvious to me because I wrote it but I can see how it could be confusing

richardtagger:
heres a demo 

bistate

src/components/Button/BistateButton.tsx
unindented: Would lines 10-14 be equivalent to this?

```
export interface BistateButtonProps extends BistateButtonProps {
  activeProps: CustomizableButtonProps;
}
```

richardtagger: Yes but we cant extend ButtonProps because its a type sadly

`An interface may only extend a class or another interface.`

unindented: Aaaah.

BrendanMcK:
> Just place it inside a fixed width container

That goes against the idea of letting the button size to content, which is what we want for i18n stuffs.

The README puts the button in a fixed size container, but IRL we'd likely not want to do that to avoid clipping/wrapping

richardtagger:
What's your suggestion for sizing?

BrendanMcK:
I don't have a good one that fits into this framework. Changing just the inner button props as needed is nice and clean; but doesn't have any clear paths to getting a stable layout over both states.

Hi @mamuso - ISTR you were looking at this? From what I remember, ways of implementing this involved various forms of layout hacking; eg collapsing to 0 height or using stacking or opacity to hide/show different content, so that the computed width would always be the larger of the two strings.

swese44:
@BrendanMcK 
```
I might want to see if we can do better for the accessible name for these. Thinking here is that from
screenreader POV, we'd like this to have accessible text of "Joined - Leave" - so that a
screenreader user searching for "Leave" can find this, regardless of visual state. I'll do a follow-up
PR for this.
```

Shouldn't this be handled by the application passing the correct `ariaLabel`?

src/components/Button/BistateButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return <Button {...props} />;
+  }
+
+  private isActive = () => this.state.isHovering || this.state.hasFocus;
+
+  private onFocus = (e: React.FocusEvent<FabricButtonType>) => {
+    if (this.props.onFocus) {
+      this.props.onFocus(e);
+    }
+    this.setState({ hasFocus: true });
```
swese44: i think these handlers should `setState` before the callback, giving the user immediate feedback in case somebody does something stupid in a callback which would cause a delay. probably not a big deal but i'd switch it

BrendanMcK:
> Shouldn't this be handled by the application passing the correct ariaLabel?

This default should be better, so that in the normal case, the app doesn't have to supply aria-label. This component could set the button's aria-label to 'Text - ActiveText' as default

src/components/Button/BistateButton.tsx
richardtagger: I dont think state reconciliation will happen until this event handler has finished so not sure it matters. Dont remember react internals enough to be sure

src/components/Button/BistateButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+  constructor(props: BistateButtonProps) {
+    super(props);
+    this.state = { ...inactiveState };
+  }
+
+  public render() {
+    const propsForState = this.isActive() ? this.props.activeProps : undefined;
+    const { children, activeProps, ...buttonProps } = this.props;
+    const props: ButtonProps = {
+      fullWidth: true,
```
swese44: if you look at `<Button>`, @unindented implemented its loading state so the `text` is hidden from view/screenreaders but still rendered, ensuring the button width doesn't change when its text is replaced with a Spinner.

maybe `<Button>` could accept a prop like `longestAlternateText` which gets rendered in the same hidden manner to push the button to a max width? feels like an awkward API though... but we'd want to avoid setting a hard width on the button based on state because of perf, and also to allow the application to responsively change a button's width based on content or layout via `fullWidth: true`

richardtagger: > maybe <Button> could accept a prop like longestAlternateText which gets rendered in the same hidden manner to push the button to a max width

How does a developer know what the 'longest' text is. Long is long in DOM terms - you'd have to stick it in the DOM to measure....... iii is less long than WWW

swese44: in this case, BistateButton would pass either `text` or `activeProps.text` as Button's `text`, and pass the unused one as Button's `longestAlternateText`.

not suggesting this is a great API, just that the concept might work to push the button's width out to its max possible width without having the application try to hardcode widths. ideally BistateButton could handle this behind the scenes without surfacing a public `longestAlternateText` prop on Button...

richardtagger: i don't follow. You can't know what the widest button will be (active or inactive) until its rendered to the dom and you measure

BrendanMcK: I think PJ's naming is causing confusion here - I had the same issues initially.

I thin the suggestion is really for a 'secondarySizingText' property. Idea is that the button renders its regular content as normal, and then renders this text in some hidden way (clipped, transparent, etc). The longest of the two wins out as far as HTML's sizing goes.

richardtagger: > Idea is that the button renders its regular content as normal, and then renders this text in some hidden way (clipped, transparent, etc). The longest of the two wins out as far as HTML's sizing goes

Sounds like putting it into the DOM to measure, then?

BrendanMcK: Not really to measure (which suggests you'd do something with the results after); but just putting both strings in the DOM so that the button naturally gets the widest width of the two strings.

Net result is something like:

    <button>
        <div aria-hidden=true opacity=0 height=0>Inactive string</div>
        <div>Active String</div>
    </button>

(attrs to illustrate the concept only, apply salt as needed.)

richardtagger: ick. ok

BrendanMcK: Ick indeed! That's the sort of hackery I was referring to in my comment above. Didn't know we were already doing similar for the loading button though - so there's at least another legit use case for this sort of thing.

(Almost forgot - visibility:hidden is great for this sort of layout thing; it hides things visually and from a11y, but still does layout - so better than opacity hacking.)

richardtagger: Here's the least hacky version I could come up with https://github.com/Microsoft/YamUI/pull/290/commits/c71b633cdd055376575ce445155ca02e26c3118d

Thoughts?

richardtagger: (we might want to change children to widthPadder or some such, IDK. I dont have an informed opinion)

BrendanMcK:
Thinking more about the aria labelled stuffs - eh, maybe it's fine to have the app provide the aria-label.

Issue here is that for join/follow etc buttons, we typically want the screenreader text to be explicit and include the entity name. So in that case the app will want to provide an aria-label anyway.

So typical code might be something like:

    const leaveJoinProps = {
      text: 'Joined',
      icon: CheckMark,
      activeProps: {
        text: 'Leave',
        icon: SignOut,
      },
      ariaLabel: 'Joined - Leave Bug Reports'
    };

Thing is we'd just one one overall ariaLabel, not separate for the two states. And want to avoid something as verbose as 'Joined Bug Reports - Leave Bug Reports'.

Wondering if we should then *require* an ariaLabel for these controls?

richardtagger:
The icon also affects width - how would that fit into this secondaryText picture?

swese44:
icons are all the same width. i don't think we have the use case where active state has an icon and inactive state does not, so i'd just assume BistateButton would have an icon in both states or in neither and not worry about it

richardtagger:
Done

richardtagger:
@mamuso can I get your review here too please

src/components/Button/Button.css
```diff
@@ -88,3 +88,9 @@
 .y-button--icon-wrapper-right {
   padding-left: 4px;
 }
+
+.y-button--secondary-text {
+  visibility: hidden;
+  overflow: hidden;
+  height: 0px;
```
swese44: if this doesn't work across browsers we may need to do `height: 1px;` and `margin-top: -1px;`

src/components/Button/Button.tsx
```diff
@@ -195,11 +201,21 @@ export default class Button extends React.Component<ButtonProps> {
...
+      buttonIcon && <FixedGridColumn fixed={true}>{buttonIcon}</FixedGridColumn>;
+
     return (
       <Block textSize={textSize} className={className}>
-        {iconPosition === 'left' && buttonIcon}
-        {text}
-        {iconPosition === 'right' && buttonIcon}
+        <FixedGridRow gutterSize={GutterSize.NONE}>
+          {leftIconColumn}
+          <FixedGridColumn>
```
swese44: i was thinking this `FixedGridColumn` would also need to be `fixed` to prevent it from pushing out wider, but if the outer `display:flex Button` prevents that then we're good. may want to double check in other browsers like IE/Edge

src/components/Button/Button.css
swese44: this could also just be `height: 0;` but postcss will handle that

richardtagger:
> not sure how many times we are going to have a significant difference

With other languages this is probably common, I know that some languages dont have a convenient way to say 'Unlike' for example, so we need weird long copy. Happy to defer to your judgement here.

swese44:
@mamuso i think we're waiting on your guidance here. should we make the icon center with the text or should we ship as-is? thanks!

richardtagger:
Updated

swese44:
looks like we need to update the visual diff screenshot

`✖ Screenshots config/styleguide-visual/reference/Button_bistate_button_follow_unfollow_desktop.new.png and config/styleguide-visual/reference/Button_bistate_button_follow_unfollow_desktop.png differ in 13183 pixels
`

src/components/Button/Button.tsx
```diff
@@ -195,11 +201,19 @@ export default class Button extends React.Component<ButtonProps> {
 
     const className = this.isLoading() ? 'y-button--content__hidden' : '';
 
+    const leftIconColumn = iconPosition === 'left' && buttonIcon;
```
swese44: i'd remove `column` from the variable names since these aren't columns now

richardtagger:
updated visuals

mamuso:
NICE!

PR #295, opened 2018-04-11
WIP: Focus outlines: PR for initial review only.

BrendanMcK:
Work in progress adding focus outline support to be consistent with Fabric.
   
- add utility classes in focus-outline.css that manage the actual CSS work,
including only displaying when Fabric's .is-focusVisible is set.
    
- Update Button to use these as-is,
    
- Update Clickable and NavigationLink to use these, and also take
 display and focus options for fine tuining depending on whether
 these contain block or inline content. Also updated .md files
  with examples.
    
Outstanding work items:
- test coverage
- image diffs
- bumping rev etc.
- fix issue with NavigationLink containing Image - this ends up with extra
     inline line padding at bottom. Need to add display=block option to Image?

### DIFF
--- a/src/components/NavigationLink/NavigationLink.tsx
+++ b/src/components/NavigationLink/NavigationLink.tsx
@@ -1,7 +1,7 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 import '../../yamui';
 import * as React from 'react';
-import { join } from '../../util/classNames';
+import { camelCaseToDashed, join } from '../../util/classNames';
 import { NestableBaseComponentProps } from '../../util/BaseComponent/props';
 import './NavigationLink.css';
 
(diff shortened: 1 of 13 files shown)


src/components/NavigationLink/NavigationLink.tsx
```diff
@@ -45,16 +59,29 @@ export default class NavigationLink extends React.Component<NavigationLinkProps>
...
-    const { className, unstyled } = this.props;
+    const { className, unstyled, focus, display } = this.props;
 
     const classes: string[] = ['y-navigationLink'];
     if (unstyled) {
       classes.push('y-navigationLink__unstyled');
     }
+
+    if (focus) {
+      classes.push(`y-focus-${camelCaseToDashed(focus)}`);
```
unindented: Is it worth doing `camelCaseToDashed`? We're using camelCase in classnames in this repo.

unindented: It may actually be a BEM modifier, so you could just do `y-focus--${focus}`

BrendanMcK:
Closing in favor of mergeStyles-based outline handling.

PR #300, opened 2018-04-21
bump office-ui-fabric-react to 5.85.0

chrismohr:
In preparation to allow custom description's on TextField.
https://github.com/OfficeDev/office-ui-fabric-react/pull/4588

## Changes

1. Avatar - Persona initials changed line-height from 70px to 72px. Added overrided to fix.
https://github.com/OfficeDev/office-ui-fabric-react/pull/4382/files

2. MenuButton - Fabric IconButton refactor caused a change in css order. If you look at the commit history, my first fix was to use more specific selectors.  Then instead I passed the overrides in through the `styles` prop on IconButton.  I think this will generally be more maintainable, but don't feel great about the hard-coded color strings. Feedback?

3. Hovercard - .ms-Callout now has a 1px border, when we expected 0.  I've added an !important overrride here.

4. Various snapshot changes.

### DIFF
--- a/src/components/MenuButton/MenuButton.tsx
+++ b/src/components/MenuButton/MenuButton.tsx
@@ -9,6 +9,7 @@ import {
 } from 'office-ui-fabric-react/lib/ContextualMenu';
 import { join } from '../../util/classNames';
 import { BaseComponentProps } from '../../util/BaseComponent/props';
+import Colors from '../../util/colors';
 import { IconButton, IButtonStyles } from 'office-ui-fabric-react/lib/Button';
 import { IconSize, BaseIcon } from '../Icon';
 import MoreIcon from '../Icon/icons/More';
@@ -154,20 +155,20 @@ export default class MenuButton extends React.Component<MenuButtonProps> {
         opacity: 0.7,
         width: `${this.props.iconSize}px`,
         height: `${this.props.iconSize}px`,
-        color: '#495361',
+        color: Colors.button,
       },
       rootHovered: {
         backgroundColor: 'transparent',
-        color: '#386cbb',
+        color: Colors.buttonHovered,
         opacity: 1,
       },
       rootPressed: {
         backgroundColor: 'transparent',
-        color: '#386cbb',
+        color: Colors.buttonPressed,
         opacity: 1,
       },
       rootExpanded: {
-        color: '#386cbb',
+        color: Colors.buttonExpanded,
         opacity: 1,
       },
     };
--- a/src/components/MenuButton/MenuButton.tsx
+++ b/src/components/MenuButton/MenuButton.tsx
@@ -9,7 +9,7 @@ import {
 } from 'office-ui-fabric-react/lib/ContextualMenu';
 import { join } from '../../util/classNames';
 import { BaseComponentProps } from '../../util/BaseComponent/props';
-import { IconButton } from 'office-ui-fabric-react/lib/Button';
+import { IconButton, IButtonStyles } from 'office-ui-fabric-react/lib/Button';
 import { IconSize, BaseIcon } from '../Icon';
 import MoreIcon from '../Icon/icons/More';
 import MenuButtonItem from './MenuButtonItem';
@@ -97,6 +97,7 @@ export default class MenuButton extends React.Component<MenuButtonProps> {
   public render() {
     return (
       <IconButton
+        styles={this.getStyles()}
         ariaLabel={this.props.ariaLabel}
         menuProps={this.getMenuProps()}
         onRenderIcon={this.getIcon}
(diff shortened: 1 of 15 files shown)


src/components/MenuButton/MenuButton.tsx
```diff
@@ -144,4 +145,31 @@ export default class MenuButton extends React.Component<MenuButtonProps> {
...
+
+  private getStyles = (): IButtonStyles => {
+    return {
+      root: {
+        border: 0,
+        padding: 0,
+        opacity: 0.7,
+        width: `${this.props.iconSize}px`,
+        height: `${this.props.iconSize}px`,
+        color: '#495361',
```
unindented: It may be time to create a TS file with all these colors, and have it live side-by-side with the CSS one. Having these colors sprinkled in TS code doesn't look great.

src/components/MenuButton/MenuButton.tsx
```diff
@@ -144,4 +145,31 @@ export default class MenuButton extends React.Component<MenuButtonProps> {
...
+        backgroundColor: 'transparent',
+        color: '#386cbb',
+        opacity: 1,
+      },
+      rootExpanded: {
+        color: '#386cbb',
+        opacity: 1,
+      },
+    };
+  };
```
unindented: You will be generating a new object every time this function gets called, which means you'll need to re-render this component every time.

unindented: Looks like `getMenuProps` had the same problem, so this isn't making it worse.

src/css/variables/borders.css
```diff
@@ -6,4 +6,5 @@ $borderRadius__round: 50%;
   box-shadow: 0 0 12px rgba(0, 0, 0, 0.2) !important;
   border: none;
   border-radius: 1px;
+  border-width: 0px !important;
```
unindented: No need for units when 0, just `border-width: 0 !important`.

src/components/MenuButton/MenuButton.tsx
swese44: i agree, i don't know what color that is just looking at the PR :(

src/components/MenuButton/MenuButton.tsx
swese44: the styles can change if certain props change, so i guess we could cache the styles object and update it when props change. seems gross. another option could be to make the entire styles object static, specifying variable values like root width and height on specific root classes and passing the correct root className in at each size?

src/util/fabric/theme.ts
```diff
@@ -1,65 +1,42 @@
...
       fontSize: '2.2rem',
       lineHeight: '2.8rem',
     },
     xxLarge: {
-      ...fontStyles,
       fontSize: '2.4rem',
       lineHeight: '2.8rem',
     },
-    superLarge: fontStyles,
-    mega: fontStyles,
```
unindented: Nice.

chrismohr: Weird thing is that it looks like these fontStyle colors had no effect.

swese44: these did override the default typography element colors. are visual diffs still passing?

src/util/colors.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+/*
+ * Use for component level overrides
+ */
+const colors = {
+  button: deathMetal,
+  buttonHovered: river,
+  buttonPressed: river,
+  buttonExpanded: river,
```
swese44: in CSS we abstracted all these colors into pseudo "theme" variables which components would map to their own rules. i think we're getting too specific here, i think we should be mapping our un-understandable blue/stream/pond/deathmetal color names to contextual theme names like Fabric does.

swese44: otherwise this `colors` object will turn into an enormous object with lots of duplication

src/util/fabric/theme.ts
chrismohr: Yes, the visual diffs still passed.  I manually inspected some elements, but had trouble finding the rule being applied anywhere.

Looking closer now at `Block`, I see the color is being overridden by a merge style, so instead of `#343A41`  it is `rgb(51, 51, 51)`.

I'll look a little closer at this, and see if our visual diff picks up the difference.

src/util/colors.ts
chrismohr: @swese44 Makes sense.  So maybe `textColorSecondary` and `linkColor`?

chrismohr: the word `color` is probably a bit redundant. How about this?

const colors = {
  textSecondary: deathMetal,
  link: river,
};

PR #302, opened 2018-04-24
allow textfield description to accept jsx

chrismohr:
This will allow us to add an "edit" link for the group name field's description.


src/components/BaseTextField/index.tsx
```diff
@@ -71,11 +71,34 @@ export interface BaseTextFieldProps extends BaseComponentProps {
   onMouseLeave?: ((event: React.MouseEvent<HTMLInputElement | HTMLTextAreaElement>) => void);
 }
 
+/**
+ * This is a hack to allow us to accept jsx elements in the description prop.
```
swese44: do we plan on updating the Fabric component to allow onRenderDescription without a description?

src/components/BaseTextField/index.tsx
```diff
@@ -71,11 +71,34 @@ export interface BaseTextFieldProps extends BaseComponentProps {
...
+ *
+ * We depend on onRenderDescription to display the description. However, the fabric
+ * textfield will only invoke it when it has a truthy description or errorMessage prop.
+ * Therefore we pass an arbitrary truthy string. It's value is only used to invoke
+ * onRenderDescription, and is otherwise ignored.
+ */
+const descriptionPresent = (props: BaseTextFieldProps) => {
+  if (props.description === undefined) {
+    return undefined;
```
swese44: should this return false? as-is the function's return type is `true | undefined`

src/components/BaseTextField/index.tsx
```diff
@@ -71,11 +71,34 @@ export interface BaseTextFieldProps extends BaseComponentProps {
...
+  return () => <span className="y-base-text-field--description">{description}</span>;
+};
+
 export function getBaseTextFieldProps<T extends BaseTextFieldProps>(props: T): FabricTextFieldProps {
   return {
     value: props.value,
     label: props.label,
-    description: props.errorMessage ? undefined : props.description,
+    description: descriptionPresent(props),
+    onRenderDescription: getOnRenderDescription(props.description, props.errorMessage),
```
swese44: getOnRenderDescription generates one of 2 new functions on each render cycle. instead of passing getOnRenderDescription, can we pass this.onRenderDescription which either returns null or the span with description?

src/components/BaseTextField/index.tsx
chrismohr: The type is actually `string | undefined`, which is what it needs to be.

Maybe using the string `'true'` here is confusing, do you think something like `'ignored truthy string'` be better?

src/components/BaseTextField/index.tsx
chrismohr: This is a bit of a challenge, since we need the `description` prop passed to YamUI's TextField, which we aren't able to pass into fabric's TextField because of the type mismatch.

src/components/BaseTextField/index.tsx
chrismohr: I'd like to, but would prefer not to block this change on it.

swese44: sounds good, can you add an issue to track it?

src/components/BaseTextField/index.tsx
swese44: oh duh, that's a string not a bool. eh it's fine, this can go away when we get a chance to update fabric

src/components/BaseTextField/index.tsx
swese44: ah i see

src/components/BaseTextField/index.tsx
chrismohr: https://github.com/OfficeDev/office-ui-fabric-react/issues/4669

PR #306, opened 2018-04-26
add focus() to TextField

chrismohr:
add focus() to TextField

## Pull request checklist


src/components/TextField/TextField.tsx
```diff
@@ -24,11 +25,21 @@ export interface TextFieldProps extends BaseTextFieldProps, DebouncedOnChangePro
   underlined?: boolean;
 }
 
+interface Focusable {
+  /**
+   * Sets focus to the input, and places cursor at the end of the value.
+   * @public
+   */
+  focus(): void;
+}
```
chrismohr: This was a shot at surfacing this function in the stlyeguidist docs, but didn't work.  Any ideas?

src/util/DebouncedOnChange/index.tsx
```diff
@@ -18,13 +18,25 @@ export interface DebouncedOnChangeProps {
...
+   * Optional callback to access the component interface. Use this instead of ref for accessing
+   * the public methods and properties of the component.
+   */
+  componentRef?(a: any): void;
 }
 
 export interface DebouncedOnChangePrivateProps {
   /**
-   * Used to pass both onChange and debouncedOnChange to the contained component.
+   * Private: Used to pass both onChange and debouncedOnChange to the contained component.
```
unindented: The whole interface is pretty clear about it's private status, maybe we don't need to reiterate in the description?

src/util/DebouncedOnChange/index.tsx
```diff
@@ -18,13 +18,25 @@ export interface DebouncedOnChangeProps {
...
    */
   unifiedOnChange?: ((newValue: any) => void);
+
+  /**
+   * Private: Optional callback to access the component.  Used to pass componentRef
```
unindented: Same here.

src/components/TextField/TextField.tsx
chrismohr: I got the idea from here: https://react-styleguidist.js.org/docs/documenting.html#public-methods

src/components/TextField/TextField.test.tsx
```diff
@@ -16,6 +16,19 @@ describe('<TextField />', () => {
...
       expect(component).toMatchSnapshot();
     });
+
+    it('focus', () => {
+      const instance = component.instance() as any;
+      const fabricTextFieldElement = { focus: jest.fn(), setSelectionRange: jest.fn() };
+      const focusSpy = jest.spyOn(fabricTextFieldElement, 'focus');
+      const setSelectionRangeSpy = jest.spyOn(fabricTextFieldElement, 'setSelectionRange');
+
+      instance.fabricTextFieldElement = fabricTextFieldElement;
```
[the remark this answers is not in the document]
chrismohr: It started to feel a bit dirtier, so I backed out. It would need to be in a separate file, since the snapshot tests depend on the real fabric text field, then I felt a bit challenged about making a fake ref function that would return fake ref.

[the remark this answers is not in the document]
chrismohr: I took another run at it. This time using a full mount and spying on functions on the prototype of the fabric component. I think this should address your concerns about brittleness. 

I spent some time trying `jest.mock('office-ui-fabric-react/lib/TextField', ...)` but failed to find a way to construct a fake react component that took props. (Since the YamUI Textfield depends on Fabrics componentRef, the fake would need to replicate it's behavior. )

src/components/TextField/TextField.tsx
chrismohr: Maybe it's because of https://github.com/styleguidist/react-docgen-typescript.  I don't see anything about functions there.

src/components/TextField/TextField.test.tsx
```diff
@@ -48,3 +50,42 @@ describe('<TextField />', () => {
     });
   });
 });
+
+describe('TextField#focus', () => {
```
swese44: can we put this inside the main describe block instead of starting a new outer block?

src/components/TextField/TextField.test.tsx
```diff
@@ -48,3 +50,42 @@ describe('<TextField />', () => {
...
+  const setRef = (i: any) => (ref = i);
+
+  describe('with default options', () => {
+    beforeEach(() => {
+      mount(<TextField componentRef={setRef} />);
+    });
+
+    it('focus', () => {
+      const focusSpy = jest.spyOn(FabricTextField.prototype, 'focus');
+      const setSelectionRangeSpy = jest.spyOn(FabricTextField.prototype, 'setSelectionRange');
```
swese44: these 2 declarations are repeated in another block below. i'd declare them in the top-level describe block. then i'd make this `it` block a `describe` block which explains the verb/action it's taking, then nest the expectation in an `it` block. maybe something like this:

```
describe('with componentRef', () => {
  let ref: ITextField | undefined;
  let focusSpy: jest.Mock;
  let setSelectionRangeSpy: jest.Mock;

  const setRef = (node: ITextField) => (ref = node);

  describe('when focused', () => {
    const focusSpy = jest.spyOn(FabricTextField.prototype, 'focus');
    ref.focus();
  });

  it('focuses the text input', () => {
    expect(focusSpy).toHaveBeenCalled();
  });

  // Multiple expectations should be out here in their own `it` block
});
```

src/components/TextField/TextField.tsx
```diff
@@ -37,11 +40,26 @@ class TextField extends React.Component<TextFieldProps & DebouncedOnChangePrivat
...
+    if (el) {
+      const valueLength = this.props.value ? this.props.value.length : 0;
+
+      el.focus();
+      el.setSelectionRange(valueLength, valueLength);
+    }
+  };
+
+  private getRef = (ref: ITextField | null) => {
+    this.fabricTextFieldElement = ref ? ref : undefined;
```
swese44: this could be slightly simplified to

`this.fabricTextFieldElement = ref || undefined;`

src/components/TextField/TextField.tsx
```diff
@@ -29,6 +30,8 @@ export interface TextFieldProps extends BaseTextFieldProps, DebouncedOnChangePro
  * a single line of text. The text displays on the screen in a simple, uniform format.
  */
 class TextField extends React.Component<TextFieldProps & DebouncedOnChangePrivateProps> {
+  private fabricTextFieldElement: ITextField | undefined;
```
swese44: i believe it's possible for Fabric to return `null` in the ref callback, unless that recently changed? if it could be null, this should be `ITextField | null` and then we should set `this.fabricTextFieldElement = null;` in the constructor

src/components/TextField/TextField.tsx
```diff
@@ -37,11 +40,26 @@ class TextField extends React.Component<TextFieldProps & DebouncedOnChangePrivat
...
     );
   }
 
+  public focus = () => {
+    const el = this.fabricTextFieldElement;
+    if (el) {
+      const valueLength = this.props.value ? this.props.value.length : 0;
+
+      el.focus();
+      el.setSelectionRange(valueLength, valueLength);
```
swese44: the browser's default behavior with `element.focus()` is to place the cursor at the user's previous location. is there a reason we're manually moving the cursor to the end instead? not saying this is the wrong choice, just wondering what the reason is and if we should ask for UX guidance here

src/components/TextField/TextField.tsx
```diff
@@ -37,11 +40,26 @@ class TextField extends React.Component<TextFieldProps & DebouncedOnChangePrivat
         suffix={this.props.suffix}
         underlined={this.props.underlined}
         onChanged={this.props.unifiedOnChange}
+        componentRef={this.getRef}
         {...getBaseTextFieldProps(this.props)}
       />
     );
   }
 
+  public focus = () => {
```
swese44: do we need to also duplicate these methods and ref on `TextArea`? we may want to implement an internal `BaseTextField` component which restricts the Fabric props to only what our `TextField` and `TextArea` need (and handle CSS styling), then our `TextField` and `TextArea` components would be super thin wrappers which just manage their list of props

src/components/TextField/TextField.tsx
chrismohr: In the case I'm implementing, we're going from a disabled to enabled at the same time, and at the end seemed like the right place. 

I can see not always wanting this behavior though, in which case we may also want to expose something like setSelectionRange, or a simplified version of it. 

I've posted on yammer asking for UX guidance.

src/components/TextField/TextField.tsx
chrismohr: > i believe it's possible for Fabric to return null in the ref callback
Still is, in this case `getRef` converts the `null` to `undefined`.

Your suggestion is how I originally had it, but then refactored to this, preferring not to have the constructor if we didn't need it. Do you see an advantage to changing it back?

src/components/TextField/TextField.tsx
chrismohr: See the other comment below. The subtle difference is the return types.  Your suggestion returns `ITextField | null` and the existing one is `ITextField | undefined`.

src/components/TextField/TextField.test.tsx
chrismohr: Yup, looks good.  I'll do that.

src/components/TextField/TextField.tsx
chrismohr: I don't have a plan to enable this on TextArea.  Would you be comfortable if we deferred this until needed?

src/components/TextField/TextField.test.tsx
chrismohr: done

src/components/TextField/TextField.test.tsx
chrismohr: done

src/components/TextField/TextField.tsx
swese44: yep that's fine

src/components/TextField/TextField.tsx
swese44: React itself returns a component/node or `null`, so if we replace `null` with `undefined` we're just doing something nonstandard. other YamUI components also use `null` so i'd prefer consistency, but it's not a big deal since this property is private to this component.

src/components/TextField/TextField.test.tsx
```diff
@@ -49,43 +49,62 @@ describe('<TextField />', () => {
...
-      const setSelectionRangeSpy = jest.spyOn(FabricTextField.prototype, 'setSelectionRange');
+      beforeEach(() => {
+        focusSpy = jest.spyOn(FabricTextField.prototype, 'focus');
+        setSelectionRangeSpy = jest.spyOn(FabricTextField.prototype, 'setSelectionRange');
+        mount(<TextField componentRef={setRef} />);
+      });
 
-      ref.focus();
+      it('calls focus', () => {
+        ref.focus();
```
swese44: this action should also be up in the beforeEach. otherwise the describe block 'when focused' is sort of a lie. and you wouldn't have to repeat the focus action for each expectation

src/components/TextField/TextField.test.tsx
```diff
@@ -49,43 +49,62 @@ describe('<TextField />', () => {
...
+      let setSelectionRangeSpy: jest.SpyInstance;
+
+      beforeEach(() => {
+        focusSpy = jest.spyOn(FabricTextField.prototype, 'focus');
+        setSelectionRangeSpy = jest.spyOn(FabricTextField.prototype, 'setSelectionRange');
+      });
 
-      ref.focus();
+      it('calls focus', () => {
+        ref.focus();
```
swese44: also these two

src/components/TextField/TextField.tsx
```diff
@@ -37,11 +40,26 @@ class TextField extends React.Component<TextFieldProps & DebouncedOnChangePrivat
...
+    const el = this.fabricTextFieldElement;
+    if (el) {
+      const valueLength = this.props.value ? this.props.value.length : 0;
+
+      el.focus();
+      el.setSelectionRange(valueLength, valueLength);
+    }
+  };
+
+  private getRef = (ref: ITextField | null) => {
```
swese44: this should probably be named `setRef` since it's a setter and not returning anything

src/components/TextField/TextField.test.tsx
chrismohr: I agree. Thanks for catching this.

src/components/TextField/TextField.test.tsx
chrismohr: done

src/components/TextField/TextField.tsx
chrismohr: good eye

PR #307, opened 2018-05-01
Add @vanschyndel as Designer approver

swese44:


swese44:
also added @cainhimself

PR #310, opened 2018-05-02
Spinner alignment and color

swese44:
- Adds a new "metadata" color to Spinner to support upcoming PreviewCard
- Changes default alignment to left, with option to align centered

* [x] Component `README.md` file is up-to-date.
* [x] Component is unit tested.

PR #311, opened 2018-05-02
Support click event on Box component

### LINKED ISSUE #220: Box border pushes content in 1px

swese44:
- Adds support for click event on Box's outer div (with button role for accessibility)
- Fixes 1px too much margin on Box due to 1px border - fixes #220
- Fixes possible Block width issues due to its 100% width rule. Now only gets width: 100% when nested within a Clickable.
- Removes unnecessary examples and visual diffs for Box; they'll be covered by the upcoming PreviewCard component.

## Pull request checklist


swese44:
i'll set the base branch to master once #310 is merged

PR #312, opened 2018-05-02
Adding TextField props to support EditableText and PreviewCard

### LINKED ISSUE #219: Textfield Underlined Bugs
I've spotted a few bugs when revisiting Textfield Underlined. Since we're also going to use them in Search Box we should fix them and make sure both components are working correctly.

- Font-size and height are wrong compared to design spec (check attachment)
- On Error case, the error hint is 1px vertically misaligned
- There is no focus case, when the input is active it should have the focus state (blue line)
- Make sure that focus case remains red when the error is active too (as it happens on base textfield)

Let me know if something is not clear enough!

image

swese44:
- Adds `maxLength` and `keyDown` props to TextField
- Fixes the dimensions of the smaller underlined version of TextField. fixes #219 

## Pull request checklist


swese44:
i'll set the base branch to master once #311 is merged

src/components/TextField/TextField.tsx
```diff
@@ -1,13 +1,16 @@
...
 import * as React from 'react';
+import { join } from '../../util/classNames';
 import { getBaseTextFieldProps, BaseTextFieldProps } from '../BaseTextField';
 import DebouncedOnChange, { DebouncedOnChangeProps, DebouncedOnChangePrivateProps } from '../../util/DebouncedOnChange';
 import { TextField as FabricTextField, ITextField } from 'office-ui-fabric-react/lib/TextField';
 
 import '../BaseTextField/BaseTextField.css';
 import './TextField.css';
 
+export { ITextField };
```
unindented: Why is this necessary?

swese44: it's used by EditableText when storing the underlying TextField ref (in the next PR). but actually i can probably use our TextField type directly since @chrismohr implemented the debounced wrapper to expose it. i'll update this 👍

swese44: removed. in the next PR #313 i added an interface for TextField to implement which declares our `focus` method.

PR #313, opened 2018-05-02
EditableText component, necessary to support PreviewCard

swese44:
## Pull request checklist

* [x] Component `README.md` file is up-to-date.
* [x] Component is unit tested.

PR #314, opened 2018-05-02
PreviewCard component

swese44:
## Pull request checklist

* [ ] Component is unit tested.


swese44:
i'll set the base branch to master once #313 is merged

swese44:
@cainhimself just to verify, the "X" close button should have different spacing from the top than from the right?

swese44:
@cainhimself thanks! This component uses the Spinner component as-is, so if it needs different spacing between the spinning icon and its text we should address that in another PR directly in the Spinner component

Issue #274, opened 2018-03-06
Support tree-shaking by adding a es2015 build target


swese44:
Fabric has multiple build targets, we could do something similar. For reference:

Here's the PR which introduced the es2015 build: https://github.com/OfficeDev/office-ui-fabric-react/pull/3260

Here's the `ts` build task which has multiple targets:
https://github.com/OfficeDev/office-ui-fabric-react/blob/72b103d989212d8c3e7976e6dfc285c17356d9be/scripts/tasks/ts.js

swese44:
UPDATE

we've learned that the underlying Fabric library will soon release 6.x, replacing its default `/lib` export directory with the previously-named /`lib-es2015` which uses es6 modules. instead of adding an alternate build target in YamUI, we've decided to follow suit and export es6 modules only to `/dist`.

apps consuming YamUI which use webpack 3.x will need to update their webpack build config accordingly to transpile YamUI's es6 modules. apps on webpack 4.x will be able to benefit from tree-shaking.

fyi @a-tarasyuk

chrismohr:
Done #324

PR #316, opened 2018-05-05
Support tree-shaking by adding a es2015 build target

a-tarasyuk:
[Microsoft/YamUI#274](https://github.com/Microsoft/YamUI/issues/274)


swese44:
hey thanks @a-tarasyuk. there are also CSS build steps which use postcss to build global and component CSS artifacts in the build directory, these steps would also need to copy their contents to the new build directory.

hey @unindented, any ideas for validating the build output in each directory?

a-tarasyuk:
@swese44 Thanks, will add.

a-tarasyuk:
@swese44 Which `node`/`npm` versions do you use? I see that `package-lock.json` contains many changes, however, only one package was installed. When I switch to `master` and run `npm i` -  `package-lock.json` also updates, I suppose we use different node/npm versions.

swese44:
@a-tarasyuk i'm on node 8.8.1 and npm 5.4.2 at the moment. it looks like jest (at least in the YamUI repo) has issues on node 10.0.0, i haven't investigated why yet

swese44:
ah bad timing :) we've decided to support es6 modules directly instead of an alternate build target:
https://github.com/Microsoft/YamUI/issues/274#issuecomment-387909239

@a-tarasyuk we'd love to still have you as a contributor. if you want to close this PR and open a new one with a commit to support es6 modules in the default `tsconfig.json` file i can build on that and integrate the newer version of Fabric

a-tarasyuk:
@swese44 ok, thanks

Issue #317, opened 2018-05-07
Add wrapper for PeoplePicker


machineloop:
For our CC bar on feed publishers, we'll need to use PeoplePicker from Fabric, what do we need to do to get started? I already worked on a version of PeoplePicker for selecting groups in Sharepoint v1. @swese44 @richardtagger Any objections to creating a new pass-through component in YamUI for PeoplePicker?

<img width="1288" alt="screen shot 2018-05-07 at 1 06 27 pm" src="https://user-images.githubusercontent.com/3682072/39722408-7c37cd3c-51f7-11e8-9e53-cc26b049a297.png">

richardtagger:
Nope the pass through is what we agreed - I'd start with something like what we do with Callout. If we want to wrap the component to narrow down or improve the API later that's fine

PR #321, opened 2018-05-09
feat(UserPicker): Add UserPicker wrapper for PeoplePicker from Fabric

machineloop:
As per #317 , adds wrapper pass-through component for PeoplePicker from Fabric, called UserPicker. Needed a few style tweaks for color contrast ratios.

The UserPicker is a thin wrapper of Fabric's NormalPeoplePicker is used to select one or more entities, such as people or groups. Entry points for PeoplePickers are typically specialized TextField-like input fields known as a "well", which are used to search for recipients from a list. When a recipient is selected from the list, it is added to the well as a specialized Persona that can be interacted with or removed. Clicking on a Persona from the well should invoke a PersonaCard or open a profile pane for that recipient.

Only required prop is onResolveSuggestions which can be a promise or function which returns a list of the IPersona or other supplied type.

## Pull request checklist

screen shot 2018-05-25 at 2 50 02 pm


src/components/PeoplePicker/PeoplePicker.test.tsx
```diff
@@ -0,0 +1,19 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import * as React from 'react';
+import { shallow, ShallowWrapper } from 'enzyme';
+import PeoplePicker, { IBasePickerSuggestionsProps } from '.';
+
+describe('<PeoplePicker />', () => {
+  let component: ShallowWrapper<IBasePickerSuggestionsProps>;
+
+  describe('with additional className', () => {
```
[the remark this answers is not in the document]
machineloop: Thanks! Added more descriptive describe

richardtagger:
@swese44 mind taking a look at this?

src/components/PeoplePicker/PeoplePicker.md
```diff
@@ -0,0 +1,32 @@
+### Notes for use
+
+PeoplePicker is simply NormalPeoplePicker [exported from Fabric](https://developer.microsoft.com/en-us/fabric#/components/peoplepicker).
```
unindented: Why mention the underlying implementation? Maybe describe when it should be used instead.

src/components/PeoplePicker/PeoplePicker.md
```diff
@@ -0,0 +1,32 @@
...
+
+
+### Examples
+
+Basic usage:
+
+```js { "props": { "data-description": "NormalPeoplePicker example" } }
```
unindented: Same here, I wouldn't mention `NormalPeoplePicker` in the description, cause if we change the underlying implementation we'll have to change this, and then the screenshot will have a different name and we won't be able to compare the before and after easily.

src/components/PeoplePicker/PeoplePicker.tsx
```diff
@@ -0,0 +1,10 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import '../../yamui';
+import { IBasePickerSuggestionsProps, NormalPeoplePicker } from 'office-ui-fabric-react/lib/Pickers';
+
+/**
+ * PeoplePicker is simply NormalPeoplePicker
+ * [exported from Fabric](https://developer.microsoft.com/en-us/fabric#/components/peoplepicker).
```
unindented: Same thing here, I wouldn't mention the underlying implementation in a comment (it's obvious from the `import`), but rather explain what it does and when it should be used.

src/components/PeoplePicker/PeoplePicker.md
machineloop: The primary reason I didn't add a lengthy description is the link to the Fabric docs contains that information and specifics may become stale here. In the Callout pass-through component, we provide similar [limited description](https://github.com/Microsoft/YamUI/blob/master/src/components/Callout/Callout.md#notes-for-use). But I'll add additional general information about what the component should be used.

swese44: since we're implementing this component backward (Engineering-focused instead of Design/UX-focused) think it makes sense to link out to Fabric's docs. but we should remove this when we dial it in and limit its props

src/components/PeoplePicker/PeoplePicker.md
swese44: agree, this data-description could just be "basic"

swese44:
Styleguidist has a limitation where it can't build component docs for components directly exported from a 3p library. i don't see visual diff reference images, does this component even render in the documentation app?

machineloop:
Yeah, you're right @swese44 , it doesn't render in documentation app. In Yammer thread you mentioned we should just go ahead and create more of a thin wrapper component instead of the direct pass-through done here, preferred way forward?

richardtagger:
is this ready to review again @machineloop

machineloop:
@swese44 , I need to add a few style overrides to fix issues with the vanilla Fabric component, but can't find any great docs on the proper way to do it, and the naming conventions for the classes.

For example, here's text that's rendered too light on hover:
screen shot 2018-05-16 at 1 02 07 pm

machineloop:
@richardtagger, see above comment with screenshot, need to add CSS style overrides to tweak

machineloop:
@richardtagger, @swese44  Could use your eyes again for review, thanks!

PR #323, opened 2018-05-10
Update EditableText to move focus back to the text when it exits edit…

BrendanMcK:
… mode.

Also reworked how focusing works - it's no longer a special case for TextField:
- new Focsuable interface added to src/util
- FocusableComponentProps added to BaseComponent/props
- pattern established where focusable components can expose their
  Focusable interface via a focusableRef property
- Replaced the TextFieldComponent code with above Focusable
- added similar to Clickable
- now EditableText can use same techique to focus either the edit
  or the clickable text as appropriate.

## Pull request checklist

### DIFF
--- a/src/components/EditableText/EditableText.test.tsx
+++ b/src/components/EditableText/EditableText.test.tsx
@@ -2,8 +2,13 @@
 import * as React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
 import Clickable from '../Clickable';
-import TextField, { TextFieldComponent } from '../TextField';
+import TextField from '../TextField';
 import EditableText, { EditableTextProps } from './index';
+import { Focusable } from '../../util/Focusable';
+
+type SetFocusableRef = (focusable: Focusable) => void;
+type EventHandler = (event: any) => void;
+type ChangeHandler = (text: string) => void;
 
 describe('<EditableText />', () => {
   let component: ShallowWrapper<EditableTextProps>;
@@ -99,13 +104,13 @@ describe('<EditableText />', () => {
 
       describe('and text is changed', () => {
         beforeEach(() => {
-          const onChange: (text: string) => void = component.find(TextField).prop('onChange') as (text: string) => void;
+          const onChange = component.find(TextField).prop('onChange') as ChangeHandler;
           onChange('new text');
         });
 
         describe('and ENTER is pressed', () => {
           beforeEach(() => {
-            const onKeyDown: (event: any) => void = component.find(TextField).prop('onKeyDown') as (event: any) => void;
+            const onKeyDown = component.find(TextField).prop('onKeyDown') as EventHandler;
             onKeyDown({ which: 13, preventDefault: jest.fn() });
             component.update();
           });
@@ -125,7 +130,7 @@ describe('<EditableText />', () => {
 
         describe('and ESC is pressed', () => {
           beforeEach(() => {
-            const onKeyDown: (event: any) => void = component.find(TextField).prop('onKeyDown') as (event: any) => void;
+            const onKeyDown = component.find(TextField).prop('onKeyDown') as EventHandler;
             onKeyDown({ which: 27, preventDefault: jest.fn() });
             component.update();
           });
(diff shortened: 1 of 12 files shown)


src/components/EditableText/EditableText.test.tsx
```diff
@@ -240,39 +210,62 @@ describe('<EditableText />', () => {
...
+        clickableFocusable = { focus: jest.fn() };
+        const setClickableFocusable = component.find(Clickable).prop('focusableRef') as SetFocusableRef;
+        setClickableFocusable(clickableFocusable);
+
+        // Force the component open...
+        component.find(Clickable).simulate('click', fakeEvent);
+
+        // Now we can hook up the focusable ref for the TextArea...
+        textAreaFocusable = { focus: jest.fn() };
+        const setTextAreaFocusable = component.find(TextField).prop('focusableRef') as SetFocusableRef;
```
unindented: Comment and variable names say `TextArea` but component is actually `TextField`. Is this intentional?

src/components/EditableText/EditableText.tsx
```diff
@@ -122,17 +131,27 @@ export default class EditableText extends React.Component<EditableTextProps, Edi
...
   private setTextFieldFocus() {
-    if (this.textFieldRef) {
-      this.textFieldRef.focus();
+    if (this.textFieldFocusable) {
+      this.textFieldFocusable.focus();
+    }
+  }
+
+  private setClickableFocus() {
+    if (this.clickableFocusable) {
```
swese44: looks like the "else" case doesn't have test covereage here :(

src/components/EditableText/EditableText.test.tsx
```diff
@@ -99,13 +104,13 @@ describe('<EditableText />', () => {
 
       describe('and text is changed', () => {
         beforeEach(() => {
-          const onChange: (text: string) => void = component.find(TextField).prop('onChange') as (text: string) => void;
+          const onChange = component.find(TextField).prop('onChange') as ChangeHandler;
```
swese44: i think line 95 above this was necessary to test the "else" test case for focusing textfield, you'd probably have to do the same to get full coverage on the Clickable focus

src/components/TextField/TextField.tsx
```diff
@@ -26,23 +27,19 @@ export interface TextFieldProps extends BaseTextFieldProps, DebouncedOnChangePro
...
-  focus(): void;
-}
-
 /**
  * The TextField component enables a user to type text into an app. It's used to capture
  * a single line of text. The text displays on the screen in a simple, uniform format.
  */
-class TextField extends React.Component<TextFieldProps> implements TextFieldComponent {
+class TextField extends React.Component<TextFieldProps> implements Focusable {
   private fabricTextFieldRef: ITextField | null;
```
swese44: could we also use the `createRef` API here?

src/components/EditableText/EditableText.tsx
```diff
@@ -55,11 +56,13 @@ export interface EditableTextState {
...
 export default class EditableText extends React.Component<EditableTextProps, EditableTextState> {
-  private textFieldRef: TextFieldComponent | null;
+  private textFieldFocusable: Focusable | null;
+  private clickableFocusable: Focusable | null;
 
   constructor(props: EditableTextProps) {
     super(props);
-    this.textFieldRef = null;
+    this.textFieldFocusable = null;
+    this.clickableFocusable = null;
```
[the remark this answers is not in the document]
BrendanMcK: My take is that these are not general purpose refs though; they're single purpose things - that just happen to use the callback pattern to set them. So clickableFocusable is a thing that lets you focus the Clickable.

src/components/TextField/TextField.tsx
BrendanMcK: Unsure - createRef is designed for use with React refs, which get you a ref the the underlying React element. This thing is more of a Fabric convention that gets you a helper interface. It might work, but I figured I'd leave it alone since it's not the same mechanism.

PR #324, opened 2018-05-10
Change module/target to es2015

a-tarasyuk:
#274


swese44:
FYI to other reviewers, we can build on this PR when upgrade to Fabric 6.x

PR #325, opened 2018-05-10
Release version 7.0.0

BrendanMcK:
Bumping version. Since we're changing the name of one of the props (componentRef -> focusableRef), it's a breaking change by semver rules.


unindented:
@swese44 any other breaking changes we want to put out as part of 7.0.0?

swese44:
i think we're good here. thanks @BrendanMcK!

PR #330, opened 2018-05-11
Adding ariaLabel to NavigationLink

swese44:
## Pull request checklist

### DIFF
--- a/src/components/NavigationLink/NavigationLink.tsx
+++ b/src/components/NavigationLink/NavigationLink.tsx
@@ -17,7 +17,12 @@ export interface NavigationLinkProps extends NestableBaseComponentProps {
   newWindow?: boolean;
 
   /**
-   * Title or description of the linked document for the anchor tag.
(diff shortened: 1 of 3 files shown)


src/components/NavigationLink/NavigationLink.tsx
```diff
@@ -17,7 +17,12 @@ export interface NavigationLinkProps extends NestableBaseComponentProps {
   newWindow?: boolean;
 
   /**
-   * Title or description of the linked document for the anchor tag.
+   * Title or description of the linked document for screenreaders.
+   */
+  ariaLabel?: string;
```
richardtagger: under what conditions would you not provide a label? When you're providing a title instead or?

swese44: title shouldn't really be used, it was added for backward compatibility with old tooltips which read the title attribute. ariaLabel will replace all visible content in the link with its own text value

PR #333, opened 2018-05-14
Button pressed state

### LINKED ISSUE #133: Button pressed state is missing
We should include the pressed state for both primary and secondary buttons.

http://zpl.io/aMLv6Y3

chrismohr:
Correct colors for button pressed state.
fixes #133 

First commit, is a css only change. 

Then in the second commit I moved the color css to js. I have some doubts about this direction though, since we now have button styles in 2 files instead of 1, adding potential confusion.  Thoughts?


src/util/colors.ts
```diff
@@ -61,6 +62,13 @@ export const theme: IPartialTheme = {
 const colors = {
   textSecondary: deathMetal,
   link: river,
+  river,
+  lake,
+  bay,
+  white,
+  newBlue,
+  indieRock,
+  pond,
```
chrismohr: I considered exporting semantically named colors here, but it felt like an unnecessary and potentially confusing extra layer of abstraction.  I found that when I was in `ButtonStyles.ts`, I just wanted to know the actual color names.

src/util/colors.ts
```diff
@@ -8,14 +8,15 @@ import { IPartialTheme } from 'office-ui-fabric-react/lib/Styling';
...
 // const blue = '#2477c3';
 // const stream = '#e9eff8';
-// const pond = '#9ac3ff';
+const pond = '#9ac3ff';
 const lake = '#6c98d9';
 const river = '#386cbb';
-// const bay = '#264f8c';
+const bay = '#264f8c';
 // const ocean = '#19345d';
+const newBlue = '#2f5c9f';
```
chrismohr: I've posted a question in yammer to clarify with design if we need this new color.

src/util/colors.ts
swese44: i think this is fine for now. i wonder if it would make sense for component styles (like ButtonStyles.ts) to pull the contextual color names from `theme.palette` instead of `colors`? or maybe we can just provide additional colors in `theme.palette` and let Fabric components pick up our colors automatically?

src/components/Button/ButtonStyles.ts
```diff
@@ -0,0 +1,58 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+
+import Colors from '../../util/colors';
+import { IButtonStyles } from 'office-ui-fabric-react/lib/components/Button/Button.types';
+import { ButtonColor } from './enums';
+
+const primaryColors: IButtonStyles = {
```
swese44: instead of `primaryColors` and `secondaryColors`, maybe we could use names `primary` and `secondary` and replace all of our CSS with JS styles? that's the eventual goal, as we'll get more good stuff like RTL support and locale-specific fonts.

swese44:
cool. we'll be migrating all of our components to use JS styles like this soon. ideally we'll be able to remove all component CSS and take advantage of mergeStyles

src/util/colors.ts
chrismohr: I gave this some more thought, and I think I now agree that we don't want to spread the yammer color names into the components.  I've updated our colors to be a bit more like fabric's, with both a `palette` that has names like `themePrimary`, `neutralDark` etc. AND `semanticColors` that have `bodySubtext`, `link` etc.  Both of these, using the types defined by fabric.

For component level styles, I think we should try to use a semanticColors if the name fits well, and if not, then use a color from the palette.

src/components/Button/ButtonStyles.ts
chrismohr: Sounds good.

src/components/Button/Button.styles.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export const getStyles = (props: ButtonProps): IButtonStyles => {
+  const { size, color, status, fullWidth } = props;
+
+  const colors = color === ButtonColor.PRIMARY ? primaryColors : secondaryColors;
+  const { root, rootHovered, rootFocused, rootPressed } = colors;
+
+  return {
+    root: {
+      ...(root as {}),
```
swese44: why is the `as {}` necessary?

chrismohr: `[ts] Spread types may only be created from object types.`

src/components/Button/Button.styles.ts
```diff
@@ -0,0 +1,93 @@
...
+      textAlign: 'center',
+      display: fullWidth ? 'block' : 'inline-block',
+      position: 'relative',
+      whiteSpace: 'nowrap',
+      transition: 'all 0.05s ease-out',
+      borderRadius: Styles.borderRadiusSoft,
+      borderWidth: '1px',
+      borderStyle: 'solid',
+      boxShadow: '0 1px 1px 0 rgba(0, 0, 0, 0.05)',
+      fontWeight: '600',
```
swese44: our font weights could also be copied from our CSS variables to the JS styles in `utils/`

src/components/Button/Button.styles.ts
swese44: `root` is an object, i was just wondering why the `as {}` was necessary but not a big deal

src/util/colors.ts
```diff
@@ -43,24 +43,31 @@ const lestat = '#d1423b';
...
 };
 
 /*
- * Use for component level overrides
+ * Used globally by Fabric
  */
-const colors = {
-  textSecondary: deathMetal,
-  link: river,
+export const semanticColors: Partial<ISemanticColors> = {
```
swese44: 👍

PR #334, opened 2018-05-14
Fabric v6

unindented:
Upgrading to the latest Fabric v6 beta seems to be relatively straightforward. No visual differences, and only minor changes in tests and code. Should we go ahead, so that we can start moving style overrides to JS?

cc @swese44 @richardtagger @chrismohr


config/jest/config.json
```diff
@@ -6,9 +6,10 @@
   "modulePaths": ["src/components", "src/util"],
   "testPathIgnorePatterns": ["<rootDir>/build", "<rootDir>/dist", "<rootDir>/config", "<rootDir>/src/demos"],
   "moduleNameMapper": {
+    "office-ui-fabric-react/lib/": "office-ui-fabric-react/lib-commonjs/",
```
unindented: Jest doesn't understand ES2015 modules, so point it to CommonJS files.

config/jest/config.json
```diff
@@ -6,9 +6,10 @@
...
     "^.+\\.css$": "identity-obj-proxy"
   },
-  "snapshotSerializers": ["enzyme-to-json/serializer"],
+  "snapshotSerializers": ["enzyme-to-json/serializer", "@uifabric/jest-serializer-merge-styles"],
```
unindented: `merge-styles` dynamically generates class names that can change at any time, and they end up in snapshots, so this serializer expands them to get reliable tests: https://www.npmjs.com/package/@uifabric/jest-serializer-merge-styles

src/util/fabric/theme.test.ts
```diff
@@ -1,5 +1,7 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
-jest.mock('office-ui-fabric-react/lib/Styling');
+jest.mock('office-ui-fabric-react/lib/Styling', () => ({
+  loadTheme: jest.fn(),
+}));
```
unindented: This file now just re-exports `@uifabric/styling`, and Jest wasn't processing it properly.

config/jest/config.json
swese44: since Fabric 6.x changed the `/lib` directory to be es6 modules, don't we need to change all of our Fabric import statements to import from `/lib-commonjs`? otherwise i think we need to commit to es6 modules in this PR so consumer apps can update their builds?

unindented: Umm, let me check what the output from `tsc` looks like.

I was able to launch Storybook without changes though.

unindented: Yup, the output converts our `import`s to `require`s, but of course they still point to `/lib/`. Things work because Webpack understands how to deal with those, no extra config needed.

PR #335, opened 2018-05-15
Bunch of misc PRs before adding focus style support

BrendanMcK:
Few small-ish changes ahead of adding focus outline style supports.

Add test to button to ensure PropTypes validation test works
--

Replace FakeLink with Clickable in markdown files
-- 

FakeLink should only ever be used within a Clickable or NavigationLink - it should never be used stand-alone.

feat(block): Add block option to Image and NavigationLink
--
Image needs block to avoid adding additional line height when its contained within another block
    
Navigation needs block so that it can correctly contain block content. (CSS splits inline elements that contain block content into two, which can break some properties.)
    
Also updated Clickable.md with scenario where it wraps a block Image.

## Pull request checklist

* [ ] Component is unit tested.


src/components/Clickable/Clickable.md
```diff
@@ -30,3 +30,27 @@ Unstyled block with FakeLink:
...
+```
+
+Wrapping image:
+
+```js { "props": { "data-description": "wrapping image" } }
+const yammerLogo = 'logo.png';
+const yammerLogoDescription = 'Yammer "y" logo';
+
+<div>
+  <Clickable onClick={action('clickable clicked')} unstyled block={true}>
```
swese44: this should be `unstyled={true}`, linters don't work in .md files :/

BrendanMcK:
> could you also set <ActionLink>'s <NavigationLink> to use block={true}?

Sure! I actually had that change ready for a different set of PRs, but it works fine here too... - new commit added.

Also fixed the missing unstyled={true} items.

Issue #223, opened 2018-01-26
Tooltip has white corners


swese44:
This is only visible with a BG color other than white. We should also add an example with a dark background.

Issue #225, opened 2018-01-26
Tooltip covers its trigger content if child is a block-level element


swese44:
Tooltip wrapper element should be inline-block instead of inline.

Current:
![screen shot 2018-01-26 at 10 53 20 am](https://user-images.githubusercontent.com/471731/35455578-4a166e5e-0287-11e8-9954-d8f1f66bb0a7.png)

Expected:
![screen shot 2018-01-26 at 10 53 11 am](https://user-images.githubusercontent.com/471731/35455579-4a2ad6be-0287-11e8-8a16-a4f619c7d11c.png)

PR #338, opened 2018-05-15
Fix tooltip

chrismohr:
Fixes:
https://github.com/Microsoft/YamUI/issues/223
https://github.com/Microsoft/YamUI/issues/225

Moves styles from css to js.

Includes latest visual diffs from https://github.com/Microsoft/YamUI/pull/303 , which should be merged first.


src/util/styles.ts
```diff
@@ -5,7 +5,12 @@ import { ICSSPixelUnitRule } from '@uifabric/merge-styles/lib/IRawStyleBase';
 const borderRadiusSoft: ICSSPixelUnitRule = '2px';
 const fontWeightBold: IFontWeight = '600';
 
+const fontSizeSmall = '1.2rem';
```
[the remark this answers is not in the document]
chrismohr: Would you be alright with me tackling this in a separate PR?  Since Text and Block will be the primary consumers of these values, I'd like them to inform the types and shape we choose. 

I've started a spike here:
https://github.com/Microsoft/YamUI/pull/344

src/components/Tooltip/Tooltip.css
```diff
@@ -1,26 +1,4 @@
...
-.y-tooltip--callout > .ms-Callout-main,
-.y-tooltip--callout > .ms-Callout-beak,
-.y-tooltip--callout > .ms-Callout-beakCurtain {
-  background-color: $textColor;
-}
-
-.y-tooltip--callout > .ms-Callout-beakCurtain {
-  border-radius: $borderRadius__soft;
+.y-tooltip.ms-TooltipHost {
+  display: inline-block;
```
swese44: any chance we could get this last style in JS?

chrismohr: I've run out of ideas, so filed an issue:
https://github.com/OfficeDev/office-ui-fabric-react/issues/4892

PR #211, opened 2018-01-17
[wip] Add choice group

chrismohr:
Add choice group component based on:
- https://app.zeplin.io/project/5907d6451d2b06c6b36b2852/screen/5a0b482ef7b96e16349eb438
- https://www.yammer.com/microsoft.com/#/notes/3332927

Todo:

- [ ] Fix spacing below images (There are 12px being added to the height of the image wrapper, but haven't yet been able to find a css rule responsible.)

PR #340, opened 2018-05-16
feat(choicegroup): Initial ChoiceGroup Implementation

richardtagger:
This adds ChoiceGroup component.

Out of scope:-
* Error
* Disabled
* Image tile style

Inspired by a previous attempt https://github.com/Microsoft/YamUI/pull/211/

## Pull request checklist


src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * The label for the option.
+   */
+  label: string;
+
+  /**
+   * The sublabel for the option
+   */
+  sublabel?: string;
```
swese44: if we keep with our other similar implementations this should probably be named "text"
https://github.com/Microsoft/YamUI/blob/master/src/components/Dropdown/Dropdown.tsx#L17

actually label shoud be optional and text should be required. would be consistent with Dropdown, and looks like label isn't always included in the mockups

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Label text that will be rendered above the dropdown.
+   */
+  label?: string;
+
+  /**
+   * Intial selected option. If none is provided and selectionOptional is `false`, the first option is used.
+   */
+  initialSelection?: ChoiceGroupOption['key'];
```
swese44: i'd suggest replacing `initialSelection` with `selectedKey` and just making this a controlled component, unless you think we need it to be uncontrolled?

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+  selectionOptional?: boolean;
+
+  /**
+   * Function to call when user changes the selected item.
+   */
+  onChange(key: ChoiceGroupOption['key']): void;
+}
+
+export interface ChoiceGroupState {
+  selectedKey: ChoiceGroupOption['key'] | undefined;
```
swese44: if this is a controlled component we could remove the internal state

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+    <span className="ms-Label">
+      <Text bold={true}>{`${label}:`}</Text>
+      <Text>&nbsp;{sublabel}</Text>
+    </span>
+  );
+};
+
+const mapOption = (option: ChoiceGroupOption) => {
+  const text = option.sublabel ? '' : option.label;
+  const onRenderLabel = option.sublabel ? getLabelWithSubLabel(option.label, option.sublabel) : undefined;
```
swese44: is there a solution where the onRenderLabel method isn't creating a new function on every render cycle? if it's a private method on our ChoiceGroup component then it could access this.props

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onChange(key: ChoiceGroupOption['key']): void;
+}
+
+export interface ChoiceGroupState {
+  selectedKey: ChoiceGroupOption['key'] | undefined;
+}
+
+const getLabelWithSubLabel = (label: string, sublabel: string) => {
+  return () => (
+    <span className="ms-Label">
```
swese44: i think Fabric's class names are going away with Fabric 6.x (which @unindented is currently upgrading us to). is this class adding styling we can handle on our own?

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+ * The ChoiceGroup component, also known as radio buttons, let users select one
+ * option from two or more choices. Each option is represented by one ChoiceGroup
+ * button; a user can select only one ChoiceGroup in a button group.
+ */
+export default class ChoiceGroup extends React.Component<ChoiceGroupProps, ChoiceGroupState> {
+  constructor(props: ChoiceGroupProps) {
+    super(props);
+    const { initialSelection, selectionOptional, options } = this.props;
+    this.state = {
+      selectedKey: initialSelection ? initialSelection : selectionOptional ? undefined : options[0].key,
```
swese44: does our linting even allow nested ternary? could we break this up so it's easier to read?

src/components/ChoiceGroup/ChoiceGroup.tsx
richardtagger: I think this component should control itself if thats what you mean

src/components/ChoiceGroup/ChoiceGroup.tsx
richardtagger: If its allowed to control itself it makes it easier to use - all you need to do is provide an onchange. otherwise everything that uses it has to implement a wrapper controlling the state

src/components/ChoiceGroup/ChoiceGroup.tsx
richardtagger: could probably make it a class method

src/components/ChoiceGroup/ChoiceGroup.tsx
richardtagger: it does. I can break it up

src/components/ChoiceGroup/ChoiceGroup.tsx
richardtagger: Ok - i will bring this inline with Dropdown

src/components/ChoiceGroup/ChoiceGroup.tsx
swese44: each app/feature rendering a ChoiceGroup will already be listening to its onChange event and tracking the changed value in its own state, it would just need to pass down a `selectedKey` prop. is there a reason to separately manage the state here and even give the consumer a chance to let the two get out of sync?

richardtagger: ah because you care about the final value rather than when it changes. Ok I will think about how to change it

swese44: we built TextField and TextArea to support both controlled and uncontrolled scenarios, but all instances in yamjs implemented them as controlled by passing down a value prop. since the parent component is already tracking its state via onChange it's not any extra work for the app, it's just a little more work to build the .md example.

looking at Fabric's Choicegroup, i don't think you need to do any state management here to get the uncontrolled scenario for free. i think you can just surface both the defaultSelectedKey and selectedKey and the app can use one or the other

src/components/ChoiceGroup/ChoiceGroup.css
```diff
@@ -0,0 +1,31 @@
...
+/* radio button border */
+.y-choice-group .is-checked.ms-ChoiceField-field::before {
+  border-width: 2px;
+  border-color: $borderColor__focused;
+}
+
+/* inside radio button */
+.y-choice-group .ms-ChoiceField-field::after {
+  border-color: $choiceGroupRadioButtonCheck;
+}
```
unindented: We're moving away from CSS overrides. You can see an example of how to achieve customization without CSS here: https://github.com/Microsoft/YamUI/pull/338/files?short_path=3a4a6ef#diff-00f5ac044af8cd3e73b8070a74488b02R1

src/components/ChoiceGroup/ChoiceGroup.tsx
unindented: We shouldn't depend on any Fabric classnames.

src/css/variables/colors.css
```diff
@@ -33,3 +33,6 @@ $alertColor__error: $color__angel;
 
 $spinnerColor__light: $color__river;
 $spinnerColor__dark: $color__white;
+
+$choiceGroupBackgroundColor: $color__white;
+$choiceGroupRadioButtonCheck: $color__river;
```
unindented: We probably don't want to add more CSS variables. We want something more like this `theme.ts` file in Fabric: https://github.com/OfficeDev/office-ui-fabric-react/blob/660800656c40963e73450fddafb9d621ce089f4a/packages/styling/src/styles/theme.ts#L129

src/components/ChoiceGroup/ChoiceGroup.css
richardtagger: According to that anatomy of a component thing:

>  A component should consist of these files:
> ComponentName.styles.ts

but i cant find it for ChoiceGroup - am I missing something?

richardtagger:
Update: Spoke with @cliffkoh and he pointed me to a PR that's in flight to move ChoiceGroup to getStyles. I'll hold off on this PR until that is merged and a release is cut https://github.com/OfficeDev/office-ui-fabric-react/pull/4852

src/components/ChoiceGroup/ChoiceGroup.test.tsx
```diff
@@ -0,0 +1,106 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import * as React from 'react';
+import { shallow, ShallowWrapper } from 'enzyme';
+import ChoiceGroup, { ChoiceGroupProps } from '.';
+// prettier-ignore
+import {
+  IChoiceGroupOptionProps,
+} from 'office-ui-fabric-react/lib/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.types';
```
chrismohr: Without `prettier-ignore`, this gets put on one line and breaks our line-length rule. 
Let me know if you know a better way around it.

https://github.com/prettier/prettier/issues/2166

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,99 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import '../../yamui';
+import * as React from 'react';
+import { BaseComponentProps } from '../../util/BaseComponent/props';
+import { ChoiceGroup as FabricChoiceGroup, IChoiceGroupOption } from 'office-ui-fabric-react/lib/ChoiceGroup';
+// prettier-ignore
```
chrismohr: see comment above

src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
```diff
@@ -19,6 +19,10 @@ exports[`<Fabric /> with default options renders 1`] = `
         font-family: inherit;
       }
       button {
+        margin-bottom: 0px;
+        margin-left: 0px;
+        margin-right: 0px;
+        margin-top: 0px;
```
chrismohr: This is from the fabric upgrade.

src/components/ChoiceGroup/ChoiceGroup.test.tsx
swese44: odd

src/components/ChoiceGroup/ChoiceGroup.md
```diff
@@ -0,0 +1,21 @@
+### Examples
+
+```js { "props": { "data-example": "no selected key" } }
```
swese44: looks like we need to delete the 2 visual diffs and re-generate them

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Function to call when user changes the selected item.
+   */
+  onChange(key: ChoiceGroupOption['key']): void;
+}
+
+const getTextWithLabel = (label: string, sublabel: string) => {
+  return () => (
+    <span className="ms-Label">
```
swese44: we should remove this classname as it's likely to change or lose its styling in the new styles world. i'd just manage the markup and bold style directly

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <Text bold={true}>{`${label}:`}</Text>
+      <Text>&nbsp;{sublabel}</Text>
+    </span>
+  );
+};
+
+const mapOption = (option: ChoiceGroupOption): IChoiceGroupOptionProps => {
+  const { text, label } = option;
+  const onRenderLabel = label ? getTextWithLabel(text, label) : undefined;
```
swese44: i wouldn't make this conditional, i'd always use our onRenderLabel method and handle the bold label conditionally in it. then the markup/styling will be consistent in both cases, less chance for Fabric to break our visual diffs in the future, slightly simpler debugging

src/util/colors.ts
```diff
@@ -72,4 +72,6 @@ export const semanticColors: Partial<ISemanticColors> = {
   buttonText: palette.neutralPrimaryAlt,
   errorText: palette.red,
   link: palette.themeDark,
+  inputBackgroundCheckedHovered: palette.themePrimary,
```
swese44: :+1:

src/components/ChoiceGroup/ChoiceGroup.test.tsx
richardtagger: Given https://github.com/OfficeDev/office-ui-fabric-react/blob/90eff99fab187ee41e325fca4d55d5c5fb838392/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/index.ts 

you should be able to change it to the shorter:-

```ts
import { IChoiceGroupOptionProps } from 'office-ui-fabric-react/lib/components/ChoiceGroup/ChoiceGroupOption';
```

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,99 @@
...
+ * option from two or more choices. Each option is represented by one ChoiceGroup
+ * button; a user can select only one ChoiceGroup in a button group.
+ */
+export default class ChoiceGroup extends React.Component<ChoiceGroupProps> {
+  public render() {
+    const { label, className, options, selectedKey } = this.props;
+    return (
+      <FabricChoiceGroup
+        label={label}
+        className={join(['y-choice-group', className])}
```
richardtagger: what does this class do now?

chrismohr: Nothing, do you think we should drop it?  I kind of like having it for consumers, but don't feel strongly.

richardtagger: i dont see a reason to keep it

swese44: let's keep it for consistency since all of the YamUI components have a top-level class name to identify them, which also makes it easy to identify the markup in production. we can have a separate discussion if we want to remove these once our components are all using CSS-in-JS

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,101 @@
...
+  onChange(key: ChoiceGroupOption['key']): void;
+}
+
+const getTextWithLabel = (text: string, label?: string) => {
+  return () => {
+    if (label) {
+      return (
+        <span>
+          <Text bold={true}>{`${text}:`}</Text>
+          <Text>&nbsp;{label}</Text>
```
swese44: This may not be a dependable way to manage layout/spacing. If a single space character is fine then we should separate the 2 Text components with a JSX space: `{' '}` Otherwise if Design tells you this is a few pixels off, we'll want to remove the space and manage the entire width with styles.

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,101 @@
...
+   */
+  onChange(key: ChoiceGroupOption['key']): void;
+}
+
+const getTextWithLabel = (text: string, label?: string) => {
+  return () => {
+    if (label) {
+      return (
+        <span>
+          <Text bold={true}>{`${text}:`}</Text>
```
swese44: text and label are swapped here

src/components/ChoiceGroup/ChoiceGroup.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <Text>&nbsp;{label}</Text>
+        </span>
+      );
+    }
+    return <Text>{text}</Text>;
```
swese44: fine as-is, but this may be cleaner?

```
const labelContent = label ? <Text bold={true}>{label}:</Text> : null;
<Fragment>
  {labelContent}{' '}{text}
</Fragment>
```

src/components/ChoiceGroup/ChoiceGroup.tsx
chrismohr: Ahh, that makes much more sense. Good catch.

PR #343, opened 2018-05-16
Initial version of Pivot component

ArturoSalazarB16:
This PR creates the first version of the Pivot Component.

Out of scope:
-NotificationBadge
-Tab title capitalization

## Pull request checklist


src/components/Pivot/Pivot.tsx
```diff
@@ -0,0 +1,93 @@
...
+  selectedKey: string;
+}
+
+export default class Pivot extends React.Component<PivotProps, PivotState> {
+  constructor(props: PivotProps) {
+    super(props);
+    this.onLinkClick = this.onLinkClick.bind(this);
+
+    this.state = {
+      selectedKey: this.props.selectedKey,
```
swese44: there's a similar conversation happening in rich's ChoiceGroup PR. you shouldn't need to manage any state here, you can just let the app rendering this component pass selectedKey. the app can manage state via onChange and pass the new selectedKey value down

src/components/Pivot/Pivot.tsx
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Key of the selected pivot item at a certain point.
+   */
+  selectedKey: string;
+}
+
+export default class Pivot extends React.Component<PivotProps, PivotState> {
+  constructor(props: PivotProps) {
+    super(props);
+    this.onLinkClick = this.onLinkClick.bind(this);
```
swese44: you can just change the method on the class instead of binding in the constructor, here's an example:
https://github.com/Microsoft/YamUI/blob/master/src/components/PreviewCard/PreviewCard.tsx#L246

src/components/Pivot/Pivot.tsx
```diff
@@ -0,0 +1,93 @@
...
+      });
+    }
+  }
+
+  private getClasses() {
+    const classes: string[] = ['y-pivot'];
+    const { className, isDark } = this.props;
+
+    if (isDark) {
+      classes.push('is-dark');
```
swese44: class name should be 'y-pivot__isDark'

src/components/Pivot/Pivot.md
```diff
@@ -0,0 +1,30 @@
...
+  { text: 'ALL', key: '1' },
+  { text: 'FOLLOWING', key: '2' }
+];
+
+<Pivot onChange={action('onChange')} pivotItems={pivotItems} />
+```
+
+With Dark theme:
+
+```js { "props": { "data-description": "basic" } }
```
swese44: need to change the data-description to something like "with dark theme", this is used in the image diff filename. at the moment one of these overwrote the other so there was only one generated image diff

src/components/Pivot/Pivot.md
```diff
@@ -0,0 +1,30 @@
+### Notes for use
+
+Add here any notes on how to use the component, or delete this section.
```
swese44: you can probably just delete the "Notes for use" section, it's pretty straightforward

src/components/Pivot/Pivot.tsx
ArturoSalazarB16: Got it, I added this prop to state to keep track of the latest selected key, and only fire the 'onChange' event when the key changed, but I think I can do the comparison with props.selectedKey, right?

swese44: yep, that should work fine

ArturoSalazarB16: Alrighty, I've addressed the comments

swese44:
looks good. you can ping a designer to take a look. i think @BrendanMcK also had concerns around non-selected header title contrast ratio, not sure if that should block this PR or be opened as an issue

ArturoSalazarB16:
@vanschyndel @cainhimself can you please have a look at this PR when you get a chance?

src/components/Pivot/Pivot.css
```diff
@@ -0,0 +1,29 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+@import 'src/css/variables/colors.css';
+@import 'src/css/variables/fonts.css';
+
+.y-pivot .ms-Pivot-link {
+  @extend text__small;
+  color: $color__deathMetal;
+  font-weight: $fontWeight__bold;
+  height: 28px;
```
swese44: i'd make this 2.8rem so it grows if the user adjusts their browser's base font size

src/components/Pivot/Pivot.css
```diff
@@ -0,0 +1,29 @@
...
+
+.y-pivot .ms-Pivot-link.is-selected {
+  opacity: 1;
+}
+
+.y-pivot .ms-Pivot-link.is-selected::before {
+  border-bottom-color: $color__river;
+}
+
+.y-pivot.y-pivot__isDark .ms-Pivot-link {
```
swese44: should this be `y-pivot__overDarkBackground`?

src/components/Pivot/Pivot.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import './Pivot.css';
+
+export interface PivotItem {
+  ariaLabel?: string;
+  key: string;
+  text: string;
+}
+
+export interface PivotProps extends NestableBaseComponentProps {
```
swese44: we're not rendering children, so this should extend `BaseComponentProps` instead

src/components/Pivot/Pivot.tsx
```diff
@@ -0,0 +1,74 @@
...
+      this.props.onChange(item.props.itemKey);
+    }
+  };
+
+  private getClasses() {
+    const classes: string[] = ['y-pivot'];
+    const { className, overDarkBackground } = this.props;
+
+    if (overDarkBackground) {
+      classes.push('y-pivot__isDark');
```
swese44: should probably use `y-pivot__overDarkBackground` for consistency

ArturoSalazarB16:
Thanks for the feedback @swese44, I've addressed your comments

ArturoSalazarB16:
The requested changes are committed, so moving forward with this PR.

PR #344, opened 2018-05-16
js styles for Text and Block

chrismohr:
Establish a pattern for js styles for components that are not backed by fabric components.

- used merge-styles
- used context for a case where the Text component needs to know about the text size on Block

**! Don't be intimidated by the size, it's mostly snapshots !** 😄


src/components/Text/Text.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+  xLarge: '2.2rem',
+  xxLarge: '2.4rem',
+};
+
+export const getStyles = (props: TextProps): IRawStyle => {
+  const { size, maxWidth, bold, uppercase, color } = props;
+
+  return {
+    ...(maxWidth ? ellipsisStyle : {}),
+    ...(bold ? fontWeightBold : {}),
```
unindented: I'd check the compiled output for these spreads. It may be much more costly than the `undefined` approach you're taking below.

src/util/styles.ts
```diff
@@ -1,11 +1,13 @@
...
-import { IFontWeight } from '@uifabric/styling/lib';
-import { ICSSPixelUnitRule } from '@uifabric/merge-styles/lib/IRawStyleBase';
+import { IRawStyle } from '@uifabric/styling';
 
-const borderRadiusSoft: ICSSPixelUnitRule = '2px';
-const fontWeightBold: IFontWeight = '600';
+export const borderRadiusSoft: IRawStyle = { borderRadius: '2px' };
+export const borderRadiusRound: IRawStyle = { borderRadius: '50%' };
+export const fontWeightBold: IRawStyle = { fontWeight: '600' };
+export const fontWeightNormal: IRawStyle = { fontWeight: '600' };
```
unindented: Why export the object instead of just the value? It forces you to spread all over the place.

chrismohr: It was an artifact of a previous iteration, where it got me past a typing challenge. I've update it to export values now.

swese44: looks like error text color changed, was this on purpose?

src/components/Block/Block.styles.ts
```diff
@@ -0,0 +1,64 @@
...
+  medium: '1.2rem',
+  large: '1.6rem',
+  xLarge: '2rem',
+  xxLarge: '2.4rem',
+};
+
+const getMarginTop = (topSpacing?: GutterSize, push?: number) => {
+  if (push) {
+    // For negative push, use top margin to "pull" it up
+    return push < 0 ? `${push / 10}rem` : undefined;
```
swese44: i think the logic needs to handle the case when negative push and topSpacing are both provided:

```
const margin = topSpacing ? gutterSize[topSpacing] : 0;
if (push < 0) {
    return topSpacing + push; // adds the negative push value
}
return margin | undefined;
```

src/components/Block/Block.styles.ts
```diff
@@ -0,0 +1,64 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import { BlockProps } from './Block';
+import { ellipsisStyle } from '../../util/styles';
+import { GutterSize } from '../FixedGrid/enums';
+import { textColors, fontSizes, lineHeights } from '../Text/Text.styles';
```
swese44: these 3 mappings will be used throughout other components, should we move them to `util/styles/`?

src/components/Block/Block.styles.ts
```diff
@@ -0,0 +1,64 @@
...
+import { IRawStyle } from '@uifabric/styling';
+
+export const gutterSize: { [size in GutterSize]: string } = {
```
swese44: this will also be used by many components, maybe move to `util/styles/`?

src/components/Text/Text.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+  small: '1.2rem',
+  mediumSub: '1.4rem',
+  medium: '1.5rem',
+  large: '1.8rem',
+  xLarge: '2.2rem',
+  xxLarge: '2.4rem',
+};
+
+const getHeight = (size?: TextSize) => {
+  const heights: Sizes = {
```
swese44: i'd move this const out of the function so it doesn't get instantiated every time

src/components/Text/Text.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+  };
+};
+
+/**
+ * Inline-block elements shift vertical-align from baseline to bottom when we set
+ * overflow:hidden for ellipsis. These rules ensure ellipsis Text, at each size and nested
+ * under each Block's textSize, maintains correct vertical alignment. Increased specificity
+ * to handle the nested block case :(
+ */
+export const getInlineStyles = (props: TextProps): CSSProperties => {
```
swese44: since we're dynamically generating these styles we don't need separate inline styles directly on the element. i would just handle these directly in getStyles()

also be careful about the logic to determine values based on the incoming props. i think this is incorrect for verticalHeight, lineHeight and height -- take a close look at the deleted CSS, these values change depending on the presence of the `maxWidth` prop and value of `size` prop

src/components/Text/Text.tsx
```diff
@@ -52,50 +54,32 @@ export default class Text extends React.Component<TextProps> {
...
+        {screenreaderText !== undefined ? <ScreenReaderText>{screenreaderText}</ScreenReaderText> : null}
       </span>
     );
   }
 
   private getClasses() {
-    const { bold, className, color, maxWidth, size, uppercase } = this.props;
-
+    const { className, size, maxWidth } = this.props;
     const classes = ['y-text'];
```
swese44: on Text and Block components, you could use our `join` util instead of String.prototype.join. it ignores falsey values, so you don't need to check for the existence of this.props.classNames

```
const classes = ['y-text', className];
...
return join(classes);
```

i should update our yeoman component template to use this util instead of casses.join()

src/css/utility-classes.css
```diff
@@ -1,29 +0,0 @@
-/**
```
swese44: :+1:

src/util/styles.ts
chrismohr: It was, but I was confused, undoing it now.

src/util/styles.ts
```diff
@@ -1,11 +1,13 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
-import { IFontWeight } from '@uifabric/styling/lib';
-import { ICSSPixelUnitRule } from '@uifabric/merge-styles/lib/IRawStyleBase';
+import { IRawStyle } from '@uifabric/styling';
 
-const borderRadiusSoft: ICSSPixelUnitRule = '2px';
-const fontWeightBold: IFontWeight = '600';
+export const borderRadiusSoft = '2px';
```
swese44: what do you think about objects or enums for properties with known possible values? this might be more clear in the consuming components

```
const borderRadius = {
    soft: '2px',
    round: '50%',
};
const fontWeight = {
    normal: '400',
    bold: '600',
};
```

src/components/Block/Block.styles.ts
chrismohr: From what I can tell, the old code would let a negative push win, and ignore the topSpacing.
Are you suggesting that the right thing to do would be to add the two?  If so, that's doable, but a bit painful since the gutterSize object contains strings like '1rem'.

src/components/Block/Block.styles.ts
chrismohr: Yeah, I was a bit torn, I'll move them out and see how it looks.

src/util/styles.ts
chrismohr: Yeah, I went back and forth a bit. What's challenging about the way your proposing is that the styles of all the values get inferred as `string`, and as an example `IRawStyle` doesn't accept a `string` for fontWeight, so you'd have to cast or declare as `IFontWeight`. Which is doable, but felt more cumbersome.

src/components/Text/Text.styles.ts
swese44: ah ok these depend on maxWidth so they do look correct. but i'd rather have a getVerticalAlign function which owns all of the logic and put the value directly in getStyles() instead of putting styles in 2 different places and having specificity determine which styles to use

src/util/styles.ts
swese44: not a big deal, we can revisit if we think it'll help

src/components/Text/Text.styles.ts
chrismohr: Yeah, I'd prefer not depend on specificity here either. We were before though too. 
`Increased specificity to handle the nested block case :(`
:-( indeed.

Maybe track this as another issue? or do you have a specific idea of how to solve this another way?

chrismohr: @swese44 I just pushed up a commit that deals with this using the react context api.  Please let me know what you think. 
(I still need to fix the jest tests.)

src/components/Block/Block.test.tsx
```diff
@@ -1,24 +1,31 @@
...
+
 describe('<Block />', () => {
-  let component: ShallowWrapper<BlockProps>;
+  let component: Element;
+  let componentWithContext: ShallowWrapper<BlockProps>;
 
   describe('with default options', () => {
     beforeEach(() => {
-      component = shallow(<Block>block content</Block>);
+      componentWithContext = shallow(<Block>block content</Block>);
```
chrismohr: There are two cases that snapshot Block with the context provider, one where TextSize is set, and one without.  The rest of the time, I think we want to exclude the context from the snapshot, to avoid unnecessary churn.

src/components/Text/Text.test.tsx
```diff
@@ -1,15 +1,19 @@
...
 import * as React from 'react';
-import { shallow, ShallowWrapper } from 'enzyme';
+import { renderIntoDocument, findRenderedDOMComponentWithClass } from 'react-dom/test-utils';
+import Text, { TextColor, TextSize } from '.';
 
-import Text, { TextProps, TextColor, TextSize } from '.';
+const render = (jsx: JSX.Element) => {
+  const renderedText = renderIntoDocument(jsx) as Text;
+  return findRenderedDOMComponentWithClass(renderedText, 'y-text');
+};
```
chrismohr: This will do a full render excluding the BlockContextProvider component.

package.json
```diff
@@ -19,12 +19,14 @@
     "@types/jest": "22.2.0",
     "@types/node": "7.0.12",
     "@types/prop-types": "15.5.1",
-    "@types/react": "16.3.13",
+    "@types/react": "16.3.14",
     "@types/react-dom": "16.0.5",
+    "@types/react-test-renderer": "^16.0.1",
```
swese44: can we keep these exact versions?

src/components/Block/Block.tsx
```diff
@@ -61,10 +64,9 @@ export interface BlockProps extends NestableBaseComponentProps {
   ellipsis?: boolean;
 }
 
-interface BlockStyles {
-  marginTop?: string;
-  paddingTop?: string;
-}
+const defaultContext = { textSize: undefined as TextSize | undefined };
```
swese44: instead of using `as`, i'd create an interface for the context value

```
export interface BlockContext {
  textSize?: TextSize;
}

const defaultContext = {};
```

src/components/Block/__snapshots__/Block.test.tsx.snap
```diff
@@ -15,37 +19,64 @@ exports[`<Block /> with additional className matches its snapshot 1`] = `
...
     block content
   </div>
 </div>
 `;
 
 exports[`<Block /> with default options matches its snapshot 1`] = `
-<div
-  className="y-block"
-  style={Object {}}
+<[object Object]
```
swese44: does the latest version of jest not fix this? or are we stuck waiting on enzyme's update?

src/components/Text/Text.styles.ts
```diff
@@ -0,0 +1,49 @@
...
+  small: '-0.3rem',
+  mediumSub: '-0.5rem',
+  medium: '-0.4rem',
+  large: '-0.5rem',
+  xLarge: '-0.5rem',
+  xxLarge: '-0.4rem',
+};
+
+const getVerticalAlign = (size?: TextSize) => {
+  return size ? verticalAligns[size] : '-0.4rem';
```
swese44: this should fall back to undefined, unless maxWidth was provided then 0.4rem

src/components/Text/Text.tsx
```diff
@@ -50,52 +53,32 @@ export interface TextProps extends NestableBaseComponentProps {
  */
 export default class Text extends React.Component<TextProps> {
   public render() {
+    return <BlockContext.Consumer>{blockContext => this.getContent(blockContext.textSize)}</BlockContext.Consumer>;
```
swese44: was there a reason to move the content into getContent? maybe it's just all the red and green lines, but would it be easier to understand what was being rendered if the markup was right here

src/components/Text/Text.styles.ts
```diff
@@ -0,0 +1,49 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import { TextProps, TextSize } from './Text';
+import { ellipsisStyle, fontWeightBold, fontSizes, lineHeights, textColors, Sizes } from '../../util/styles/fonts';
+import { IRawStyle } from '@uifabric/styling';
+
+const heights: Sizes = {
```
swese44: why are heights and verticalAligns duplicated here instead of using the exported values from fonts.ts directly?

src/util/styles/fonts.ts
```diff
@@ -0,0 +1,44 @@
...
+export const textColors: { [color in TextColor]: string | undefined } = {
+  primary: semanticColors.bodyText,
+  secondary: palette.neutralPrimaryAlt,
+  metadata: palette.neutralSecondary,
+  white: palette.white,
+  error: palette.redDark,
+  disabled: semanticColors.disabledBodyText,
+};
+
+export const fontWeightBold = '600';
```
swese44: should probably consolidate font weights too

```
export const fontWeights: {
  normal: 400,
  bold: 600,
}
```

src/components/Text/Text.tsx
chrismohr: Yeah, it was to work around the `jsx-wrap-multiline` tslint rule. Am I missing a way to do this within that constraint? Is this a rule we want to keep?

src/components/Text/Text.styles.ts
chrismohr: That is the logic, but it's split between this and line 47.  I've renamed this to `getVerticalAlignForMaxWidth` to make it clearer.

src/components/Block/__snapshots__/Block.test.tsx.snap
chrismohr: Latest version of jest AND replacing enzyme's `shallow` with react's shallow renderer https://reactjs.org/docs/shallow-renderer.html

I've made the change now, but I'm good either way.

src/components/Text/Text.tsx
chrismohr: Sorry, wrong ts-lint rule, it's this one:
```
[tslint] Multiline JS expressions inside JSX are forbidden (jsx-no-multiline-js)
```

swese44: ah. we've removed that linting rule in MC since libraries are using more render props these days, we could remove it here but don't worry about it here

PR #345, opened 2018-05-16
chore: add commonjs build, for consumers that don't understand ES2015

unindented:
Jest will have to be configured like this in consuming projects:

```
  "moduleNameMapper": {
    "yamui/dist/": "yamui/dist-commonjs/"
  },
```


swese44:
also need the override for Fabric in consuming app jest config?

package.json
```diff
@@ -80,15 +80,18 @@
...
-    "build:css:global": "postcss src/css/index.css --output dist/yamui-base.css",
-    "build:css:components": "postcss \"src/components/**/*.css\" --dir \"dist/components\" --base \"src/components\"",
-    "build:ts:components": "tsc --project .",
+    "build:es2015:css:global": "postcss src/css/index.css --output dist/yamui-base.css",
+    "build:es2015:css:components": "postcss \"src/components/**/*.css\" --dir \"dist/components\" --base \"src/components\"",
+    "build:es2015:ts:components": "tsc --project tsconfig.json",
+    "build:commonjs:css:global": "postcss src/css/index.css --output dist-commonjs/yamui-base.css",
+    "build:commonjs:css:components": "postcss \"src/components/**/*.css\" --dir \"dist-commonjs/components\" --base \"src/components\"",
+    "build:commonjs:ts:components": "tsc --project tsconfig.commonjs.json",
     "build": "run-s clean build:**",
```
swese44: we could save a few seconds by just copying CSS files from dist to dist-commonjs, but not an issue

unindented:
Yeah, I think if they mount they'll need it too.

package.json
unindented: Trrrue, but then they depend on order of execution. I'll wait for @chrismohr to kill all the CSS :trollface:

PR #352, opened 2018-05-25
Add hover state reference images for tooltips

chrismohr:
Remove the existing 4 tooltip reference images in favor of new ones that actually show the tooltip.


src/components/Tooltip/Tooltip.md
```diff
@@ -5,15 +5,17 @@ You should always wrap Tooltip around a tabbable element, ensuring keyboard navi
 
 Basic usage:
 
-```js { "props": { "data-description": "basic" } }
-<Tooltip text="This is the tooltip content.">
-  <Clickable>Hover over me</Clickable>
-</Tooltip>
+```js { "props": { "data-description": "basic", "data-action-states": "[{\"action\":\"hover\",\"selector\":\".y-clickable\"}]"  } }
```
swese44: can we just use single quotes instead of escaping these?

chrismohr: We could have, but then we would have had to replace them with double quotes in `react-styleguidist-visual` since single quote are invalid json.

PR #349, opened 2018-05-18
upgrade react-styleguidist-visual

chrismohr:
Upgrades to:
- react-stylguidist 
- react-stylguidist-visual

This resulted in a new version of Chromium coming in. 

While it's good to be on the latest, unfortunately the font rendering has changed slightly and has affected most of our diffs.

I've run many Travis runs to ensure that the tests are stable.

PR #353, opened 2018-05-25
[wip] No borders in screenshots

chrismohr:
Depends on https://github.com/Microsoft/YamUI/pull/349

The borders from react styleguidist can cause churn of the reference images.   This hides the border.

For example: 
There are diffs in this PR that are failing because 1px of the border is different.
https://github.com/Microsoft/YamUI/pull/344

(The problem seems loosely related to the documentation for a component changing.)

Also, for a perf improvement (~30s),  I've prevented the props from being documented in the styleguide that gets used for the visual diffs.


chrismohr:
Merging into  #349

PR #357, opened 2018-05-29
Add visual diff images for UserPicker

swese44:
just adds diff images for the existing UserPicker component. unfortunately we don't have interactive screenshots yet so this is just the initial box :/

## Pull request checklist


swese44:
@richardtagger looks like newer versions of `npm` are using `"^x.x.x"` instead of `"x.x.x"`, i think we saw this same thing happen in MC? i can't find any documentation to confirm this though

richardtagger:
yeah - I get the same when i run npm i, So i guess it's fine?

PR #359, opened 2018-05-30
Allow fixed height on fullWidth images

swese44:
Leaving height undefined with fullWidth maintains aspect ratio (existing behavior). But now you can also set a fixed height with fullWidth.

## Pull request checklist


richardtagger:
tests then LGTM

swese44:
ya good point, covered by visual diffs but better to have a specific unit test here. added

PR #361, opened 2018-06-01
Add aria-current support to NavigationLink

BrendanMcK:
For use in left-nav and other situations where a link is styled in such a way
to indicate that it is the current item. Screenreaders will read out
'current page' when aria-current="page" is set.

## Pull request checklist

### DIFF
--- a/src/components/NavigationLink/NavigationLink.tsx
+++ b/src/components/NavigationLink/NavigationLink.tsx
@@ -22,6 +22,12 @@ export interface NavigationLinkProps extends NestableBaseComponentProps {
   newWindow?: boolean;
 
   /**
+   * If present, indicates the link refers to the current page. This is usually used in cases where the link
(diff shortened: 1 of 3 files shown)


src/components/NavigationLink/NavigationLink.tsx
```diff
@@ -21,6 +21,12 @@ export interface NavigationLinkProps extends NestableBaseComponentProps {
    */
   newWindow?: boolean;
 
+  /**
+   * If present, indicates the link refers to the current page. This is usually used in cases where the link
+   * is visually styled to indicate it is selected or is the current page, such as in a navigation list.
+   */
+  ariaCurrent?: 'page';
```
swese44: do we have any reason to allow its other values? page, step, location, date, time

BrendanMcK: Not at this time; page is the only one with a current concrete use-case.

PR #362, opened 2018-06-01
add clickable context

chrismohr:
The primary goal was to remove the last of the css for the `Block` component. 

The rule is that a `Block` needs to be 100% width, when nested within a `Clickable` component.

To achieve this, I added a clickable Context Provider. 

This led to some challenges in our snapshot tests, which I've addressed to the best of my current ability, but I suspect there is room for improvement.

### DIFF
--- a/src/components/Block/Block.test.tsx
+++ b/src/components/Block/Block.test.tsx
@@ -1,27 +1,21 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 import * as React from 'react';
-import { createRenderer as createShallowRenderer } from 'react-test-renderer/shallow';
+import { create as createRenderer, ReactTestRendererJSON } from 'react-test-renderer';
 import { renderIntoDocument, findRenderedDOMComponentWithClass } from 'react-dom/test-utils';
 import Block, { GutterSize, TextColor, TextSize } from '.';
+import Clickable from '../Clickable';
 
-const shallowWithContext = (jsx: JSX.Element) => {
-  const renderer = createShallowRenderer();
-  renderer.render(jsx);
-  return renderer.getRenderOutput();
-};
-
-const renderWithoutContext = (jsx: JSX.Element) => {
-  const renderedText = renderIntoDocument(jsx) as Block;
-  return findRenderedDOMComponentWithClass(renderedText, 'y-block');
+const render = (jsx: JSX.Element) => {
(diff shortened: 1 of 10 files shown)


src/components/Block/Block.test.tsx
```diff
@@ -1,27 +1,21 @@
...
 describe('<Block />', () => {
-  let component: Element;
-  let componentWithContext: React.ReactElement<Block>;
+  let component: ReactTestRendererJSON | null;
+  let componentWithContext: ReactTestRendererJSON | null;
 
   describe('with default options', () => {
     beforeEach(() => {
-      componentWithContext = shallowWithContext(<Block>block content</Block>);
+      componentWithContext = render(<Block>block content</Block>);
```
chrismohr: Previously this would capture the context provider, which was good for ensuring that the correct value was set.  With the addition of the clickable context consumer tag, I'm having trouble finding a away to render it in a way that still captures it.

src/components/PreviewCard/PreviewCard.test.tsx
```diff
@@ -33,47 +33,58 @@ describe('<PreviewCard />', () => {
...
-          endEditingCallback();
-          component.update();
+          const renderedPreviewCard = renderIntoDocument(jsxComponent) as PreviewCard;
+          const renderedEditableText = findRenderedComponentWithType(renderedPreviewCard, EditableText);
+          (renderedEditableText.props.onEndEditing as Function)();
+          testComponent = findRenderedComponentWithType(renderedPreviewCard, PreviewCard as React.ClassType<
+            any,
+            any,
+            React.ComponentClass<{}>
+          >).render();
```
chrismohr: I don't yet understand why this gross casting was needed.

src/components/Block/Block.test.tsx
```diff
@@ -153,11 +146,27 @@ describe('<Block />', () => {
...
+        <Clickable>
+          <Block>block content</Block>
+        </Clickable>,
+      ) as Clickable;
+    });
+
+    it('matches its snapshot', () => {
+      expect(findRenderedDOMComponentWithClass(rendered, 'y-block')).toMatchSnapshot();
+    });
+  });
```
chrismohr: This is currently a bit of an integration test.

It might be better to mock the clickable context.

swese44: if it's more work i'd just leave as-is

PR #364, opened 2018-06-02
WIP: SuggestionsListItem and Basepicker

machineloop:
WIP: Do not merge, just working out loud

### DIFF
--- a/src/components/SuggestionsList/SuggestionsListItem.tsx
+++ b/src/components/SuggestionsList/SuggestionsListItem.tsx
@@ -18,7 +18,7 @@ export interface SuggestionItem {
 }
 
 export interface SuggestionsListItemProps extends SuggestionItem, NestableBaseComponentProps {
-  searchText: string;
+  searchText?: string;
   isSelected: boolean;
   onSelect(id: string | number): void;
 }
--- a/src/components/UserPicker/UserPicker.tsx
+++ b/src/components/UserPicker/UserPicker.tsx
@@ -1,26 +1,49 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 import '../../yamui';
-import { IBasePickerProps, IBasePickerSuggestionsProps, NormalPeoplePicker } from 'office-ui-fabric-react/lib/Pickers';
+import {
+  IBasePickerProps,
+  IBasePickerSuggestionsProps,
+  BasePicker,
+  IInputProps,
+} from 'office-ui-fabric-react/lib/Pickers';
 import * as React from 'react';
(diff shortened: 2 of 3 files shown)


src/components/SuggestionsList/SuggestionsListItem.tsx
```diff
@@ -18,7 +18,7 @@ export interface SuggestionItem {
 }
 
 export interface SuggestionsListItemProps extends SuggestionItem, NestableBaseComponentProps {
-  searchText: string;
+  searchText?: string;
```
machineloop: @richardtagger Seems I'll need to make some of the features optional, like I won't need to highlight search text in order to use SuggestionsListItem in place of the IPersona based list items...

src/components/UserPicker/UserPicker.tsx
```diff
@@ -1,26 +1,49 @@
...
+        onResolveSuggestions={this.onResolveSuggestions}
         className="y-userpicker__container"
         pickerSuggestionsProps={suggestionsClass}
-        onResolveSuggestions={onResolveSuggestions}
+        onRenderSuggestionsItem={this.onRenderSuggestions}
       />
     );
   }
+  private onRenderSuggestions = (props: any, itemProps: any) => {
+    return <SuggestionsListItem isSelected={true} onSelect={this.onSelect} />;
```
machineloop: @richardtagger Hmm, will I also need to handle bridging events like current selection, or will I need to make isSelected and onSelect optional props. Should I reconsider going away from the IPersona bound approach and just map props on SuggestionItem properties to the IPersona properties...

src/components/SuggestionsList/SuggestionsListItem.tsx
richardtagger: why cant you just pass it down?

richardtagger: the current search value I mean

src/components/UserPicker/UserPicker.tsx
richardtagger: How does BasePicker work with the JSX you pass - does it wrap it all in a click handler or?

richardtagger: `isSelected` - In BasePicker how does keyboard navigation work? If FocusZone or something else is taking care of things we should make this optional

I'm fine with the IPersona approach, though the point of YamUI is to look Yammery. It might not matter by this point however

swese44:
@machineloop i'd be happy to pair up and explore this if you're getting stuck

richardtagger:
Yes please @swese44 !

swese44:
closing in favor of https://github.com/Microsoft/YamUI/pull/412

PR #365, opened 2018-06-04
Move all enums.ts files to types.ts

swese44:
just for better consistency with our main repo


.prettierignore
```diff
@@ -1,6 +1,7 @@
 *.ejs
 package.json
 package-lock.json
+config/styleguide/config.js
```
unindented: Why do we want to avoid this file?

swese44: on commit, prettier kept removing parens around an arrow function's single arg which then fails the build

swese44: updating eslint rule instead

unindented: Ah, forgot we're running eslint on this codebase.

Issue #327, opened 2018-05-10
Icon should not have relative positioning with block={true}


swese44:
Icon has `position: relative` and `top: 2px` so it can render properly within a block of text. But this makes an Icon harder to position correctly when using `display: block` (to avoid extra vertical height from `line-height`).

PR #366, opened 2018-06-05
js styles for ActionLink

chrismohr:
1. Convert ActionLink styles to js
2. Add visual diffs for ActionLink hover state


src/components/Icon/Icon.css
```diff
@@ -12,6 +12,8 @@
 
 .y-icon__isBlock {
   display: block;
+  position: static;
+  margin-top: 0.2rem;
```
chrismohr: I was torn about adding the margin here.  Unless I'm missing something the alternative was to add wrapping blocks with push=2, to many uses of Icon.

src/components/MenuButton/MenuButton.md
```diff
@@ -6,13 +6,13 @@ Context menu should be used as a menu button where you want to provide options t
 
 MenuButton with default Icon, and all applicable children properties
 
-```js { "props": { "data-description": "with default icon and all menu item types" } }
+```js { "props": { "data-description": "with default icon and all menu item types", "data-action-states": "[{\"action\":\"none\"},{\"action\":\"click\",\"selector\":\".y-menu-button\"}]" } }
```
swese44: can you do all of these interactive diffs in a separate PR? the change required to fix #327 worries me a bit, it would be nice to have the screenshots in place for the current code first. thanks!

src/components/Icon/Icon.css
swese44: this actually won't work because we're adding 2px to the overall height of the Icon now. there are some tricky bits regarding #327, i'd suggest converting CSS first and then tackling #327 later (i'm not sure what the best solution is, we can put our heads together on it later)

src/components/ActionLink/ActionLink.tsx
```diff
@@ -55,12 +55,14 @@ export type ActionLinkProps = NavigationActionLinkProps | ClickableActionLinkPro
...
+    const classNames = getClassNames();
 
     const content = (
-      <FixedGridRow gutterSize={GutterSize.SMALL} className="y-actionLink--wrapper">
+      <FixedGridRow gutterSize={GutterSize.SMALL} className={classNames.wrapper}>
         <FixedGridColumn fixed={true}>
-          <Icon size={IconSize.MEDIUM} block={true} className="y-actionLink--icon" />
+          <Block push={2}>
+            <Icon size={IconSize.MEDIUM} block={true} />
+          </Block>
```
chrismohr: I'm probably missing something here, I used a Block instead of 

```css
.y-actionLink--icon {	
  margin-top: 0.4rem;	

  /* Remove when this is addressed: https://github.com/Microsoft/YamUI/issues/327 */	
  position: static;	
  top: 0;	
}
```

swese44: i think i left it as CSS because it was clear it needed to be updated when the issue was resolved, and lighter-weight solution. if you're going to move it to markup could you keep the comment with it?

chrismohr: I tried css, but switched back again, because the style was losing out to the styles on `.y-icon`.  I presume this is because the merge-styles are being loaded before the css styles.  I've added the comment above the render function.

PR #367, opened 2018-06-05
js styles for AsideSection

chrismohr:
Convert css to js.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -163,13 +163,26 @@
       "requires": {
         "@microsoft/load-themed-styles": "^1.7.13",
         "@uifabric/merge-styles": "6.0.0-beta4",
+        "@uifabric/utilities": "6.0.0-beta4",
         "tslib": "^1.7.1"
+      },
+      "dependencies": {
+        "@uifabric/utilities": {
+          "version": "6.0.0-beta4",
+          "resolved": "https://registry.npmjs.org/@uifabric/utilities/-/utilities-6.0.0-beta4.tgz",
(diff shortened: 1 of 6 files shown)


src/components/AsideSection/AsideSection.tsx
```diff
@@ -25,15 +25,19 @@ export interface AsideSectionProps extends NestableBaseComponentProps {
  */
 export default class AsideSection extends React.Component<AsideSectionProps> {
   public render() {
+    const styles = mergeStyleSets(getStyles());
```
unindented: According to Fabric's guidelines, this could be costly if re-rendered frequently. Should we be memoizing everywhere where we use `mergeStyleSets`? https://github.com/OfficeDev/office-ui-fabric-react/tree/master/packages/merge-styles#optimizing-for-performance

chrismohr: Yeah, it's probably a good idea. I've made the change.

package-lock.json
```diff
@@ -19107,7 +19143,7 @@
     },
     "typescript": {
       "version": "2.8.3",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-2.8.3.tgz",
+      "resolved": "https://yammer.pkgs.visualstudio.com/_packaging/yammer-clients/npm/registry/typescript/-/typescript-2.8.3.tgz",
```
chrismohr: I'm assuming this is from recent changes to my .npmrc

@unindented, @swese44 have you seen this before?

I'm guessing we always want `registry.npmjs.org`?

chrismohr: Some combination of adding `registry=https://registry.npmjs.com/` to the end of my .npmrc and running `rm -rf node_modules`, `npm cache clean --force` seems to have fixed this now.

PR #368, opened 2018-06-05
Add visual reference images for MenuButton in open state

chrismohr:
This adds some interactive variations to our 4 reference images for MenuButton. 

There are 4 that now show it after being clicked.
Then 1 where an item is being hovered over, and 1 after the down key has been pressed. 

Please let me know if any other variations would be helpful.


chrismohr:
Yeah, I was thinking the same thing.  I'll do it now, so that we don't have to thrash the filenames later.

chrismohr:
@vanschyndel , @swese44 do you think it's valuable to have the closed state for these captured?  Or can we see enough of the button in the open state?

The way I'm thinking about it, every visual reference image we're capturing has an associated cost (test runtime, design review time etc.), and I just want to try and make sure they all have value too.

chrismohr:
Maybe we can capture one closed one, since the color is different, and discard the other three?

swese44:
one closed instance sounds good to me

PR #369, opened 2018-06-06
js styles for Avatar

chrismohr:


src/components/Avatar/Avatar.tsx
```diff
@@ -62,54 +61,32 @@ export interface AvatarProps extends BaseComponentProps {
...
-
-    const badge = badgeContent && <div className={`y-avatar--badge y-avatar__size-${size}--badge`}>{badgeContent}</div>;
+    const {
+      badgeContent,
+      imageUrl,
+      name,
+      badgeDescription,
+      size = AvatarSize.MEDIUM,
+      imageShouldFadeIn = false,
+      borderType = BorderType.ROUND,
```
chrismohr: replaced defaultProps with assignments to avoid untestable conditionals

src/components/Icon/Icon.css
```diff
@@ -45,3 +45,7 @@
   width: 1rem;
   top: 0.1rem;
 }
+
+.y-avatar--badge > .y-icon {
+  top: 0;
+}
```
chrismohr: Moved this here for now. I think this needs to be addressed with https://github.com/Microsoft/YamUI/issues/327

src/components/SuggestionsList/__snapshots__/SuggestionsListItem.test.tsx.snap
```diff
@@ -43,8 +41,6 @@ exports[`<SuggestionsListItem /> when isSelected=true renders as expected 1`] =
   <MediaObject
     imageContent={
       <Avatar
-        borderType="round"
-        imageShouldFadeIn={false}
```
chrismohr: I don't fully understand how having these in Avatar's default props resulted them being in the snapshot here.

swese44: that's just how snapshots work for all React components, it'll always show default props

src/components/Icon/Icon.css
swese44: sounds good. could you add a comment above it referencing #327?

src/components/Avatar/Avatar.styles.ts
```diff
@@ -0,0 +1,48 @@
...
+import { IPersonaCoinStyleProps, IPersonaCoinStyles } from 'office-ui-fabric-react/lib/PersonaCoin';
+import { AvatarSize, BorderType } from './Avatar';
+import { borders } from '../../util/styles/borders';
+
+export interface AvatarStyleProps {
+  borderType: BorderType;
+  size: AvatarSize;
+}
+
+export const getPersonaCoinGetStylesFunction = (
```
swese44: 1. should we memoize **all** getStyles functions? is it a problem to create new object instances on every render cycle?
2. can we use a better name here, maybe `getPersonalCoinSyltes`?

chrismohr: I hear what you're saying. I'm wondering if you noticed that it returns a "getStyles" function, rather than the actual style object.  We pass it as a function to the fabric component, where it get's invoked with the `IPersonaCoinStyleProps`. 

I'm unsure if memoizing would be the right thing to do here, seems to me like it should maybe be a responsibility on the fabric side.

They allude to it here, but aren't yet doing it. 
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/utilities/src/classNamesFunction.ts#L14

src/components/Avatar/Avatar.css
```diff
@@ -1,62 +0,0 @@
...
-  width: 40px;
-}
-.y-avatar__size-small {
-  height: 32px;
-  width: 32px;
-}
-.y-avatar__size-xSmall {
-  height: 24px;
-  width: 24px;
-}
```
chrismohr: Seems these rules have no impact on our diffs.  @swese44 do you remember why we have them?

src/components/Avatar/Avatar.styles.ts
chrismohr: I was just thinking about this a bit more, and I think it does make sense to memoize.  I'm going to work on this a bit more.

src/components/Avatar/Avatar.css
swese44: Fabric upgrades have broken these dimensions a few times with margins and overflows, i feel we should continue to have these here

chrismohr: Ok.  I've added them back.  Would those regressions be tough to catch with visual diffs?

swese44: they'll be caught, but we'll have to re-add these css rules to fix them

swese44: i went through the file history and i think it was the nested image component which has broken in the past, and that css has been removed after the PersonaCoin refactor. feel free to remove, sorry for the churn :(

chrismohr: No problem. I'll remove the height and width from the root again.

PR #371, opened 2018-06-07
js styles for Box

chrismohr:


src/components/Box/Box.styles.ts
```diff
@@ -0,0 +1,24 @@
...
+  hasOnClick: boolean;
+}
+
+export const getClassNames = memoizeFunction((styleProps: BoxStyleProps) => {
+  const { hasOnClick } = styleProps;
+
+  return mergeStyleSets({
+    root: {
+      borderRadius: '1px',
+      border: `1px solid ${palette.neutralDark}`,
```
swese44: question, would it make more sense to use semanticColors here instead of palette? seems like palette defines the available colors, semantic maps them to what they should be used for, and components should use these shared meaning values?

swese44: looks like Fabric components use `semanticColors` when possible, i think this should be the pattern we follow to keep our colors semantically consistent with the other Fabric components
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/Button/BaseButton.styles.ts#L30

chrismohr: Yeah, I think this might be good case for a semantic color, but I'd like to understand a bit better.

Could you tell me a bit more about how you see people interacting with these semantic color names?

Do you imagine we'd maybe expose them all the way down to consumer integrators for theming? or is it more for us, as we develop new components?

If for theming, the question for me is about granularity.  `neutralDark` may not be good enough because we maybe want to provide finer grain control for a theme consumer.   So, what is the correct subset?  

So far, my strategy when encountering colors we've specified in css is to try in order:
1.  specify a fabric palette color
2. specify a fabric semantic color
3. add style rule using a fabric semantic color if there is a name that matches really well
4. add style rule using a fabric palette color

Sounds like maybe we'd get value from adding a step between 3 and 4, which would be to create our own semantic colors when their is an obvious abstraction.

I think I like how fabric names their semantic colors a bit better than how we've been doing it so for.  For example `inputBorder` ([category name][element name][checked state]).  Rather than 
 `borderColor`? (and `borderColorSecondary` and `borderColorTertiary`? But I'd probably tweak it a bit. 

https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/styling/src/interfaces/ISemanticColors.ts

**Current Groupings**

borderColor (IndieRock, neutralDark, neutralTertiary)
- Box: border
- AsideSection: title divider
- Dropdown: title divider
- ProgressIndicator: track background
- Button: secondary border

borderColorSecondary (AltRock, borderColorSecondary, altRock)
- Dropdown - divider
- MenuButton - divider

borderColorTertiary (PopRock, neutralLighter)
- HoverCard: header divider
- SuggestionList: divider between results and spinner


**My suggested groupings:**

pageDivider
- Box: border
- AsideSection: title divider
- Dropdown: title divider

menuDivider
- Dropdown - divider
- MenuButton - divider

calloutDivider
- HoverCard: header divider
- SuggestionList: divider between results and spinner

secondaryButtonBorder
- Button: secondary border

progressIndicatorTrack
- ProgressIndicator: track background

I intentionally deviated from fabric in using the word "divider" instead of "border" since in some cases we use background color to create the division, and I think there's probably some value in not tying the name to implementation.

Also, in a perfect world, I think the names of these abstractions would be part of our "design language".   I'm guess there are already certain names for these rules/concepts that the designers are already using, and it feels like I'm just trying to reverse engineer them here....

chrismohr: I didn't see your second comment until now.  The reason I'm talking about new groupings is because I don't see an existing semantic name that works well for this PR.  The closest is maybe inputBorder, but I don't think we should use it unless it matches really well.

chrismohr: Hah... just realized that I overlooked `bodyDivider` in fabric's semantic colors, which might work well here.

swese44: i appreciate you putting a lot of thought into this but i don't have time to dive into everything at the moment, let's have this conversation in person soon. don't wait on me to merge this.

my thought was our theme mechanism would simply allow a consumer to provide their own `semanticColors` overrides, i don't think our use cases call for more than that. but if we're finding cases where a Fabric semantic color doesn't exist to fit something then we'll probably need to create a new mapping of YamUI's semantic colors which the consumer could override, and this superset would be fed into both Fabric's `semanticColors` and our own components.

i don't have as much context as you in this area yet so i'll be happy to lean on your expertise when we implement theming

chrismohr: @swese44 Thanks. It sounds like we're mostly on the same page here. Maybe the difference is that I'm currently seeing us allowing consumers to override both the `palette` and `semanticColors`.  I think this is how fabric does it, but I'm still wrapping my head around it.

PR #372, opened 2018-06-07
[spike] remove callout component

chrismohr:


src/components/Tooltip/Tooltip.styles.ts
```diff
@@ -30,20 +30,21 @@ export const getCalloutStyles = (_: TooltipProps): ICalloutContentStyles => {
     },
     root: {
       backgroundColor: 'inherit',
-      boxShadow: 'none',
       padding: '0.7rem 1.2rem 0.8rem',
       minHeight: '3.2rem',
       zIndex: 1000,
-      border: 'none',
+      ...borderStylePopOver,
```
chrismohr: TODO: Find out if tooltip should have this style.

chrismohr:
Also waiting on https://github.com/Microsoft/YamUI/pull/368 to ensure we don't regress for MenuButton

chrismohr:
This is blocked because there is currently no way of styling the Callout used by Fabric's `SuggestionListBasePicker` used by our `Picker`.

unindented:
Given that our design team is trying to move towards Fabric's styles anyways, should we stop customizing the callout?

chrismohr:
@unindented Good question. Given how busy I've perceived our design team to be.  I've been holding off bringing these decisions to them, and hoping/expecting that once they've regrouped, we'll start working more closely together and address these en masse.  You're probably more plugged in than I am though. What do you recommend?

unindented:
Yup, I think you're right. Let's wait for them to regroup.

PR #374, opened 2018-06-07
js styles for EditableText

chrismohr:
I did this in two steps.
1. Add new reference images for interactive states.
2. Convert styles.


src/components/EditableText/EditableText.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+          ':focus': { color: semanticColors.buttonText },
+        },
+      };
+
+  return mergeStyleSets({
+    root,
+    clickableText: {
+      paddingLeft: '0.2rem',
+      selectors: {
+        ':hover': { textDecoration: 'underline' },
```
swese44: would it be cleaner to create a variable for these duplicate styles and reuse it?

chrismohr:
Not sure I'm following, this component doesn't have an `onClick` prop...

chrismohr:
We talked. PJ was thinking this was the PreviewCard. So nothing needs to be done.

PR #376, opened 2018-06-08
js styles for Button

chrismohr:
- added interactive reference images
- removed a couple reference images that probably weren't adding value
- refactored component to ingest props only in the render function, so that I could set defaults there
- converted css styles to js

I'm interested in code style feedback here, since my refactoring may have been a bit heavy.  I found myself preferring just about everything inlined and reducing the number of functions and variables. Interested to hear what others think.


src/components/Button/Button.tsx
```diff
@@ -155,87 +153,78 @@ export default class Button extends React.Component<ButtonProps> {
...
-    const loadingText = (this.props as LoadingButtonProps).loadingText;
-
-    const spinnerColor = color === ButtonColor.PRIMARY ? SpinnerColor.DARK : SpinnerColor.LIGHT;
-    const spinnerSize = size === ButtonSize.SMALL ? SpinnerSize.XSMALL : SpinnerSize.SMALL;
-
-    return (
-      <span className="y-button--spinner">
-        <Spinner color={spinnerColor} size={spinnerSize} text={loadingText} hideText={true} isCentered={true} />
-      </span>
+          <div className={classNames.secondaryText}>
```
chrismohr: I'm pretty sure we can make this block conditional on the presence of `secondaryText`, but am holding off since I didn't want to make this too hard to review.

src/components/Button/Button.md
```diff
@@ -527,21 +549,15 @@ const followUnfollowProps = {
...
-Link button:
-
-```js { "props": { "data-description": "link" } }
-<Button text="I am a link" href="https://www.yammer.com" />
-```
-
 Submit button:
 
-```js { "props": { "data-description": "submit button" } }
+```js { "props": { "data-description": "submit button", "data-action-states": "[]" } }
```
chrismohr: This suppresses the visual diff.

PR #377, opened 2018-06-08
Convert yammer palette colors from hex to rgb

swese44:
This unblocks Spinner work where we need to use a Fabric util to add an alpha channel to a CSS color. `colors.ts` is the only changed file, the rest are updated snapshots


chrismohr:
LGTM

PR #378, opened 2018-06-08
Provide an addAlpha CSS color util

swese44:
- Provides an `addAlpha` util for CSS colors
- Updates the jest moduleNameMapper for Fabric, as it was rewriting pathnames to Fabric's root dir instead of the nested dir. Worked previously because we were only importing components before this, which are all exported at the root directory.
- Reverts the conversion of yammer colors from hex to rgb(), turns out it wasn't necessary. sorry for the churn 😳


src/util/colors.ts
```diff
@@ -72,3 +73,16 @@ export const semanticColors: Partial<ISemanticColors> = {
...
+/**
+ * Generates a new CSS color value with alpha transparency.
+ * @param color Any valid CSS color value
+ * @param alphaPercentage Number between 0 and 100
+ */
+export const addAlpha = (color: string, alphaPercentage: number): string => {
+  const colorObject = getColorFromString(color);
+  if (!colorObject) {
+    return color;
+  }
```
chrismohr: Probably not a big deal... but is this saying that if the `color` is invalid, we'll just return it?  If so, wondering why do that, instead of throwing an error? or letting `updateA` throw an error?

swese44: the fabric method returns undefined if it can't parse the color string. do you think YamUI should throw an error and break the page if a bad color was provided somewhere?

swese44: (honest question, i'm not sure)

chrismohr: Yeah, that's what I was thinking, but I'm open to being wrong.  I imagined colors always being provided by developers, rather than users, and as a developer I would expect feedback if my input was invalid.  I don't feel strongly about this though, was mostly just asking so that I could learn, and understand better.

swese44: @unindented what do you think?

swese44: that sounds good, follows the same pattern as Fabric's method and in production the color will just be omitted from any CSS styles. updated

Issue #375, opened 2018-06-07
Aside Section Visual Diff Test is Flakey


chrismohr:
I see this test fail roughly 1/5 times when running: `y run test:visual`

config/styleguide-visual/reference/AsideSection_basic_desktop.png

They are slightly different heights.

Original:
![image](https://user-images.githubusercontent.com/601657/41131955-04dea00a-6a73-11e8-90c7-2bcbdd6a2a74.png)

New:
![image](https://user-images.githubusercontent.com/601657/41131959-0a229f80-6a73-11e8-8c9c-2d1886962513.png)

Using slowMo to slow down puppeteer further would probably mitigate this. 

To repro more consistently, you can probably reduce the slowMo value.
https://github.com/Microsoft/YamUI/blob/454b581951d09b96835c7ac617e6f62bdb927fe5/config/styleguide-visual/config.js#L16

chrismohr:
This was fixed in #379

PR #379, opened 2018-06-08
increase viewport width from 600 to 1024px

chrismohr:
This will enable us to test the interactive states of the dropdown component.  Currently the 
The open state is rendering as side panel rather than a callout, because the viewport we set for puppeteer is so narrow that it triggers "responsive mode". 

This also allows us to better test direction hints on callouts, which were currently rendering in the center because of the viewports size.

The reference images are able to stay the same width, because I hardcoded a max-width on the style guide to match the current width.

Unfortunately, a side effect of the viewport width change is that there are some virtually undetectable changes to pixels that will require us to update about 90 of the reference images. It seems mostly related to rounded corners, shadows and Illustrations.

One image actually changed, because the direction hint is now being accepted.
`config/styleguide-visual/reference/Hovercard_alignments_hover_desktop.png`

Additionally, I've included a fix for https://github.com/Microsoft/YamUI/issues/375, by adding a `wait` to the specific example.  Fixing this, allows us to reduce `slowMo` from 40 to 10. 

This also revealed a MenuButton bug, filed here:
https://github.com/Microsoft/YamUI/issues/421


chrismohr:
Here's an example diff:
button_bistate_button_join_leave_desktop_diff_png

swese44:
looks like you had a bad merge and reverted some previous code changes

package-lock.json
```diff
@@ -441,7 +441,7 @@
     },
     "app-path": {
       "version": "2.2.0",
-      "resolved": "https://onedrive.pkgs.visualstudio.com/_packaging/odsp-npm/npm/registry/app-path/-/app-path-2.2.0.tgz",
```
chrismohr: Fixed paths here.

config/styleguide/config.visualdiff.js
```diff
@@ -7,6 +7,7 @@ module.exports = Object.assign({}, config, {
   require: config.require.concat([path.resolve(__dirname, 'assets/disable-animations.css')]),
   propsParser: () => ({}), // Faster without parsing props. They're not needed for diffs.
   theme: {
+    maxWidth: 632,
```
unindented: What does this key do?

chrismohr: It constrains the width of the styleguide, so that our screenshots stay the same width. 
yamui_v8_5_0

unindented: Ah, understood.

PR #380, opened 2018-06-08
Convert Spinner CSS to styles

swese44:
## Pull request checklist


src/components/Spinner/Spinner.test.tsx
```diff
@@ -56,7 +56,7 @@ describe('<Spinner />', () => {
     });
   });
 
-  describe('with a size', () => {
+  describe('with small size', () => {
```
chrismohr: nit: should be `xsmall`?

src/components/Spinner/Spinner.styles.ts
```diff
@@ -0,0 +1,45 @@
...
+    const circleColor = color === SpinnerColor.DARK ? theme.palette.white : theme.semanticColors.link;
+
+    return mergeStyleSets({
+      root: {
+        display: 'flex',
+        alignItems: 'center',
+        flexDirection: size === SpinnerSize.LARGE ? 'column' : 'row',
+        justifyContent: isCentered ? 'center' : undefined,
+      },
+      circle: {
```
chrismohr: Rather than setting a class on the fabric `Spinner`, did you consider passing a `styles` prop?  Looks like you can style the circle this way:
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/Spinner/Spinner.styles.tsx#L19

swese44:
the diff will be cleaner after this PR merges: https://github.com/Microsoft/YamUI/pull/378

src/components/Spinner/Spinner.styles.ts
swese44: oh nice, i'll look at that

src/components/Spinner/Spinner.test.tsx
swese44: updated

src/components/Spinner/Spinner.styles.ts
swese44: done, good call

src/components/Spinner/Spinner.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+    },
+    label: {
+      margin: labelMargins[size],
+    },
+  });
+});
+
+export const getFabricSpinnerStyles = memoizeFunction((color: SpinnerColor) => {
+  const theme = getTheme();
+  const circleColor = color === SpinnerColor.DARK ? theme.palette.white : theme.semanticColors.link;
```
swese44: hey @chrismohr, take a look at this. it pulls the current theme directly from Fabric instead of from our `<Partial>` theme colors. what do you think about this approach?

chrismohr: @swese44 I like it! It'll be nice to have all of the colors available to us from the same source of truth like this, instead of just the subset that we're overriding. 

And I also like how you're importing it from util/colors, instead of directly  because I think we'll probably want to extend the fabric theme to have some of our own semantic color names.  In which case we'll be able to wrap it pretty easily.

chrismohr: @swese44 I think I'd advocate for using a differenent color name here though.

My votes would be to use `themeDark` or create our own semantic slot called something like `spinner` and set it to `themeDark`.

Reason being, is that I think we should try to minimize potential surprises.

If I were a developer using this library, I would expect that if I changed `link` color to green, only the links would change to green.

I would be surprised if the spinner also changed color.

If instead I wanted to turn everything river colored, into green, I would override `themeDark` from the palette.

src/util/styles/gutters.ts
```diff
@@ -10,3 +15,30 @@ export const gutterSize: { [size in GutterSize]: string } = {
...
+
+const getPxGutterValue = (gutterUnits: number) => {
+  const units = (gutterUnits * gutterUnit.px).toFixed(0);
+  return `${units}px`;
+};
+
+/**
+ * Get a CSS unit length value from the number of gutter units desired.
+ * @param gutterUnits The number of 4px gutter units desired.
+ * @param growWithText Uses rem instead of px, which resizes with text if the user changes their base font-size.
```
chrismohr: `growWithText` vs `resizeWithText`

src/components/Spinner/Spinner.styles.ts
chrismohr: It just occurred to me, that we should be consuming font sizes and lineHeights in the same way.  You may have pointed this out before, and I missed it, but we've defined them in both https://github.com/Microsoft/YamUI/blob/master/src/util/styles/fonts.ts#L8-L26 and https://github.com/Microsoft/YamUI/blob/master/src/util/fabric/theme.ts#L14-L40

chrismohr:
LGTM

src/components/Spinner/Spinner.styles.ts
```diff
@@ -0,0 +1,44 @@
...
+import { SpinnerColor, SpinnerSize } from './types';
+import { getGutterValue } from '../../util/styles/gutters';
+
+export interface SpinnerClassNames {
+  root: string;
+  label: string;
+}
+
+const labelMargins = {
+  [SpinnerSize.XSMALL]: `-1px 0 0 ${getGutterValue(1.5, true)}`,
```
unindented: Gutters are not part of the theme, so we assume they won't change, right?

swese44: yes

src/components/PreviewCard/__snapshots__/PreviewCard.test.tsx.snap
```diff
@@ -116,7 +116,6 @@ exports[`<PreviewCard /> when loading matches its snapshot 1`] = `
       >
         <Spinner
           color="metadata"
-          hideText={false}
```
unindented: This was intentional?

src/util/styles/gutters.ts
```diff
@@ -1,12 +1,44 @@
...
 
+export type CSSLengthUnit = string;
+
+export const gutterUnit = {
+  px: 4,
+  rem: 0.4,
+};
+
+const getRemGutterValue = (gutterUnits: number) => {
+  const units = (gutterUnits * gutterUnit.rem).toFixed(1);
```
unindented: Was the one decimal place chosen intentionally, or more like "it'll do"? I was wondering how much error we could introduce with the lack of a second decimal when talking about bigger gutters...

src/components/PreviewCard/__snapshots__/PreviewCard.test.tsx.snap
swese44: yes, it just removed the default `false` value. still resolves to false

src/util/styles/gutters.ts
swese44: it was intentional, in our font size strategy sets 1rem to 10px. so the one decimal place on a rem is equal to 1 pixel

PR #382, opened 2018-06-11
ResizeList component

swese44:
ResizeList will render a list of items on a single line, not rendering any items which don't fit in the available width. An optional `actionItem` can be provided, which will always be rendered after the other visible items.

This will support the Facepile-ish sidebar components:

Note that the design was changed and mockups were not updated. Zeplin currently shows a 6px gutter between each list item, which does not match our 4px gutters but stretched the list content to fully fill the current sidebar width on Yammer.com. The updated Design requirements asked for standard gutter sizes with the last "action" item aligned left.

## Pull request checklist


src/components/ResizeList/ResizeList.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * ResizeList renders a items on a single row, without rendering any items which don't fit in the available width.
+ * An optional "actionItem" can be provided, which will always be rendered after the other visible items.
+ */
+export default class ResizeList extends React.Component<ResizeListProps> {
+  public render() {
+    const data: ResizeListData = {
+      count: this.props.items.length,
+      cacheKey: this.props.items.length.toString(),
```
unindented: This won't create problems if someone swaps one item with a different one?

src/components/ResizeList/ResizeList.tsx
```diff
@@ -0,0 +1,98 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import '../../yamui';
+import * as React from 'react';
+import { join } from '../../util/classNames';
+import { BaseComponentProps } from '../../util/BaseComponent/props';
+import { ResizeGroup } from 'office-ui-fabric-react/lib/ResizeGroup';
```
unindented: Ha, I didn't know this existed. Do they rely on viewport `resize` events, or do they include a `ResizeObserver` polyfill?

swese44: they're literally listening to a throttled window resize event
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/ResizeGroup/ResizeGroup.base.tsx#L342

src/components/ResizeList/ResizeList.tsx
swese44: the cache just prevents recalculation, not a re-render, so everything is fine as long as the items don't change size. it looks like that's why the Fabric component offers a remeasure() public method, i'll surface that and componentRef

unindented: Ah fair.

swese44: actually that was a bit of a rabbit hole, `remeasure` only triggers the same container remeasure that window.resize triggers. looks like Fabric's ResizeGroup doesn't expect your items to change dimensions. we don't currently have this use case, but if we do we'll need to update Fabric's component to handle it

PR #383, opened 2018-06-11
js styles for FixedGrid

chrismohr:
## Pull request checklist


src/components/FixedGrid/FixedGridRow.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+  gutterSize: GutterSize;
+}
+
+export const getClassNames = memoizeFunction((styleProps: BoxStyleProps) => {
+  const { bottomSpacing, gutterSize } = styleProps;
+
+  return mergeStyleSets({
+    root: {
+      display: 'flex',
+      marginBottom: bottomSpacing ? gutterSizes[bottomSpacing] : undefined,
```
chrismohr: When https://github.com/Microsoft/YamUI/pull/380 is merged, we can use `getGutterValue` here instead.

src/components/FixedGrid/FixedGridColumn.styles.ts
```diff
@@ -0,0 +1,40 @@
...
+import { mergeStyleSets } from '@uifabric/styling';
+import { memoizeFunction } from '@uifabric/utilities';
+
+export interface FixedGridColumnStyleProps {
+  fixed?: boolean;
+  width?: number;
+  verticalAlign: 'top' | 'middle' | 'bottom';
+}
+
+const alignItems: { [i in 'top' | 'middle' | 'bottom']: 'flex-start' | 'flex-end' | 'center' } = {
```
swese44: TypeScript is now smart enough map enum values to object keys directly. See here:
https://github.com/Microsoft/YamUI/pull/380/files#diff-633c2868943735bc3166c3960a5bcd29R13

and usage here:
https://github.com/Microsoft/YamUI/pull/380/files#diff-633c2868943735bc3166c3960a5bcd29R29


does this not work for the string literal union types?

src/components/FixedGrid/FixedGridRow.tsx
```diff
@@ -24,27 +26,24 @@ export { GutterSize };
...
-    if (bottomSpacing) {
-      classes.push(`y-fixedGridRow__bottomSpacing-${bottomSpacing}`);
-    }
-    if (className) {
-      classes.push(className);
-    }
-
-    return classes.join(' ');
-  }
+  private childrenWithClassName = (className: string) => {
```
swese44: this is pretty ugly. can we just move the margin styles to a selector on the parent row?

src/components/FixedGrid/FixedGridColumn.styles.ts
chrismohr: The trouble here was that I need the values to be typed. `alignItems` can't be a string.  Do you know if it's possible to type the values without typing the keys?

src/components/FixedGrid/FixedGridRow.tsx
chrismohr: Good idea.  I'll try.

PR #387, opened 2018-06-11
js styles for LayoutList

chrismohr:


src/components/LayoutList/LayoutList.styles.ts
```diff
@@ -0,0 +1,48 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import { gutterSize as gutterSizes } from '../../util/styles/gutters';
+import { mergeStyleSets } from '@uifabric/styling';
+import { memoizeFunction } from '@uifabric/utilities';
+import { GutterSize } from '../FixedGrid';
+
+export interface LayoutListStyleProps {
+  direction: 'horizontal' | 'vertical';
+  gutterSize?: GutterSize;
+  align?: 'left' | 'right';
```
swese44: not a big deal, but probably safer to use the component prop types directly:

```
direction: LayoutListProps['direction'];
...
```

PR #389, opened 2018-06-11
Update Fabric to 6.9.0

unindented:


unindented:
@chrismohr no guess yet. If this gets merged, I'll do another try with the safer dependencies.

PR #391, opened 2018-06-11
js styles for MessageBar

chrismohr:


src/components/MessageBar/MessageBar.styles.ts
```diff
@@ -0,0 +1,33 @@
...
+import { memoizeFunction } from '@uifabric/utilities';
+import { MessageBarType } from './types';
+
+export interface MessageBarStyleProps {
+  type: MessageBarType;
+}
+
+const borderColors = {
+  [MessageBarType.WARNING]: semanticColors.warningBackground,
+  [MessageBarType.ERROR]: semanticColors.errorBackground,
```
chrismohr: What do you think of using these semantic names?   I have a tiny hesitation that `background` is misleading, but I think I can live with it.

swese44: looks good to me

swese44: we can adjust later if it makes sense

PR #394, opened 2018-06-11
js styles for ScreenreaderText

chrismohr:
Since these style rules were previously `!important` I chose to put them inline instead of on a merge-style class, so they would be less likely to get overridden. 

I think this should be fine since we won't need any of these rules to change for rtl.


src/components/ScreenreaderText/ScreenreaderText.tsx
```diff
@@ -3,10 +3,23 @@ import '../../yamui';
 import * as React from 'react';
 import { join } from '../../util/classNames';
 import { NestableBaseComponentProps as ScreenreaderTextProps } from '../../util/BaseComponent/props';
-import './ScreenreaderText.css';
 
 export { ScreenreaderTextProps };
 
+/* These styles are set inline to reduce the liklihood of being overridden. Similar to !important */
+const inlineStyles: React.CSSProperties = {
```
swese44: i would check with @BrendanMcK here. why not just keep `!important` to know we won't be changing any behavior, and use getStyles for consistency?

chrismohr: I assumed that merge-styles didn't allow `!important`, but I think I'm probably wrong, since it seems to allow you to set strings for everything.  I'll revisit this in a bit.

chrismohr: `clip`, `position` and `overflow` will not accept a string type, so I don't know how to set `!important` on those.  @swese44 any ideas?

src/components/ScreenreaderText/ScreenreaderText.tsx
```diff
@@ -3,10 +3,23 @@ import '../../yamui';
...
+  display: 'block',
+  padding: 0,
+  border: 0,
+  clip: 'rect(0 0 0 0)',
+  position: 'absolute',
+  margin: '-1px',
+  height: '1px',
+  width: '1px',
```
chrismohr: Adding `!important` to these rules would prevent them from being added.

src/components/ScreenreaderText/ScreenreaderText.tsx
unindented: How is `!important` needed in these places? Was it to protect from a dev overriding these properties through CSS? But that's not even going to be possible now, right?

chrismohr: Yeah, I don't think it should be necessary here anymore.

PR #400, opened 2018-06-12
js styles for SuggestionsList

chrismohr:


src/components/SuggestionsList/SuggestionListItem.styles.ts
```diff
@@ -0,0 +1,34 @@
...
+  isSelected: boolean;
+}
+
+const getMemoizedClassNames = memoizeFunction((styleProps, theme) => {
+  const { isSelected } = styleProps;
+
+  return mergeStyleSets({
+    root: {
+      cursor: 'pointer',
+      padding: '0.4rem 1.2rem 0.8rem 1.2rem',
```
swese44: can we use `getGutterValue()` here instead?

```
padding: `${getGutterValue(1, true) ${getGutterValue(3, true) ${getGutterValue(2, true)}`,
```

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -41,97 +40,63 @@ export interface SuggestionsListProps extends NestableBaseComponentProps {
...
 
     return (
-      <div className={join(classNames)}>
-        {results}
-        {status}
+      <div className={join(['y-suggestions-list', 'y-hc-border', className, classNames.root])}>
+        {hasResults && (
+          <ul className={classNames.resultList}>
+            {groupedItems.map((group: SuggestionItemGroupProps) => (
+              <li key={group.title}>
```
swese44: hrmm, i know this isn't your change but i would use a vertical `<LayoutList`> here with `<LayoutListItem>`s instead of a plain 'ul' and 'li's. especially since we are probably not going to include our baseline CSS bundle in the SharePoint app, default styling for `ul` and `li` elements often have typography list styles applied instead of div-style like we do.

PR #403, opened 2018-06-13
js styles for Icon

chrismohr:
Possibly too much for one PR:
- convert styles from css
- always get fontSizes and lineHeights from theme object

### DIFF
--- a/src/components/Text/Text.styles.ts
+++ b/src/components/Text/Text.styles.ts
@@ -6,10 +6,6 @@ import { mergeStyleSets } from '@uifabric/styling';
 import { ellipsisStyle, fontWeights, textColors, verticalAligns } from '../../util/styles/fonts';
 import { getTheme } from '../../util/colors';
 
-const getVerticalAlignForMaxWidth = (size?: TextSize) => {
-  return size ? verticalAligns[size] : '-0.4rem';
-};
-
 const iconSizeForSize = {
   [TextSize.XXLARGE]: '2.4rem',
   [TextSize.XLARGE]: '2.2rem',
@@ -20,8 +16,8 @@ const iconSizeForSize = {
   [TextSize.XSMALL]: '1.0rem',
 };
 
-const getMemoizedClassNames = memoizeFunction((styleProps: TextProps & { contextTextSize?: TextSize }, theme) => {
-  const { size, maxWidth, bold, uppercase, color, contextTextSize } = styleProps;
+const getMemoizedClassNames = memoizeFunction((styleProps: TextProps, theme) => {
+  const { size, maxWidth, bold, uppercase, color } = styleProps;
   const font = size ? theme.fonts[size === TextSize.MEDIUM_SUB ? 'smallPlus' : size] : undefined;
 
   return mergeStyleSets({
@@ -35,7 +31,7 @@ const getMemoizedClassNames = memoizeFunction((styleProps: TextProps & { context
       color: color ? textColors[color] : undefined,
       maxWidth: maxWidth || undefined,
       height: maxWidth && font ? font.lineHeight : undefined,
-      verticalAlign: maxWidth ? getVerticalAlignForMaxWidth(size || contextTextSize) : undefined,
+      verticalAlign: maxWidth ? '-0.4rem' : undefined,
       selectors: {
         '&.y-text .y-icon': {
           top: size === TextSize.XSMALL || size === TextSize.SMALL ? '0.1rem' : undefined,
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -66,13 +66,13 @@ export default class Text extends React.Component<TextProps> {
   }
 
   private getClasses() {
-    const { className, size, maxWidth } = this.props;
+    const { className, size, maxWidth, bold, uppercase, color } = this.props;
     return join([
       'y-text',
       size ? `y-textSize-${size}` : '',
       maxWidth ? 'y-text__ellipsis' : '',
       className,
-      getClassNames(this.props).root,
+      getClassNames({ size, maxWidth, bold, uppercase, color }).root,
     ]);
   }
 }
--- a/src/components/Text/__snapshots__/Text.test.tsx.snap
+++ b/src/components/Text/__snapshots__/Text.test.tsx.snap
@@ -102,7 +102,7 @@ exports[`<Text /> with maxWidth and size matches its snapshot 1`] = `
         max-width: 500px;
         overflow: hidden;
         text-overflow: ellipsis;
-        vertical-align: -0.5rem;
+        vertical-align: -0.4rem;
         white-space: nowrap;
       }
       & .y-icon {


src/components/Block/Block.tsx
```diff
@@ -83,15 +83,19 @@ export default class Block extends React.Component<BlockProps> {
...
 
     return (
-      <ClickableContext.Consumer>
-        {clickableContext => (
-          <BlockContext.Provider value={{ textSize }}>
-            <div className={this.getClasses(clickableContext.withinClickable)}>
-              <div className={`y-block--inner ${mergeStyles(getInnerStyles(this.props))}`}>{children}</div>
-            </div>
-          </BlockContext.Provider>
+      <TextSizeContext.Consumer>
```
chrismohr: Added consumer here, so that the context could be passed along, if `textSize` is not set in the current instance.

src/components/Icon/BaseIcon.tsx
```diff
@@ -27,44 +29,23 @@ export interface IconProps extends BaseComponentProps {
...
+    return (
+      <TextSizeContext.Consumer>
+        {textSizeContext => {
+          const textSize = textSizeContext.textSize;
+          const { block, className, size, color } = this.props;
+          const classNames = getClassNames({ block, size, color, textSize });
+
+          return this.renderIcon(join(['y-icon', className, classNames.root]));
+        }}
+      </TextSizeContext.Consumer>
```
chrismohr: I chose to bring the `render` function to the BaseIcon to consume `TextSizeContext` rather than putting it in the template, because I thought it would be a bit easier to change/maintain.  

The `render` function that was defined in the template has been renamed to `renderIcon`.

src/components/Icon/Icon.test.tsx
```diff
@@ -59,20 +58,4 @@ describe('<Icon />', () => {
...
-
-    // tslint:disable-next-line:mocha-no-side-effect-code
-    Object.keys(allExportedIcons).forEach(name => {
-      beforeEach(() => {
-        Icon = (allExportedIcons as any)[name];
-        component = shallow(<Icon />);
-      });
-
-      it('has a 240x240 viewbox', () => {
-        expect(component.getElement().props.viewBox).toBe('0 0 240 240');
```
chrismohr: I deleted this for now because the change broke it, but think I'll probably have to fix it and add it back for coverage.

src/util/styles/fonts.ts
```diff
@@ -5,26 +5,6 @@ import { IRawStyle, IFontWeight } from '@uifabric/styling';
...
-export const lineHeights: Sizes = {
-  xSmall: '1.6rem',
-  small: '1.6rem',
-  mediumSub: '2.0rem',
-  medium: '2.0rem',
-  large: '2.4rem',
-  xLarge: '2.8rem',
-  xxLarge: '2.8rem',
-};
-
```
chrismohr: Since we're always getting these from the `theme` we don't need them here anymore.

src/components/Block/Block.styles.ts
```diff
@@ -17,13 +18,15 @@ const getMarginTop = (topSpacing?: GutterSize, push?: number) => {
 
 export const getStyles = (props: BlockProps & { withinClickable: boolean }): IRawStyle => {
   const { push, textSize, topSpacing, bottomSpacing, textAlign, textColor, withinClickable } = props;
+  const theme = getTheme();
+  const font = textSize ? theme.fonts[textSize === TextSize.MEDIUM_SUB ? 'smallPlus' : textSize] : undefined;
```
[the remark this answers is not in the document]
swese44: would this match Fabric's naming? the only issue is that this would be a breaking change, so we'd need to version-bump YamJS and MC and fix all instances

src/components/Avatar/Avatar.styles.ts
```diff
@@ -41,6 +41,12 @@ export const getClassNames = memoizeFunction((size: AvatarSize) =>
       border: `solid 2px ${palette.white}`,
       background: palette.white,
       color: palette.themeDark,
+      selectors: {
+        /* Remove when this is addressed: https://github.com/Microsoft/YamUI/issues/327 */
+        '> .y-icon': {
```
swese44: maybe we should remove our uses of Context API and all the backflips in Jest unit tests and just use selectors like this...

chrismohr: Good idea.  I'll try it.  It would be nice to back out of the context api stuff, it's been a bit painful.

chrismohr: Context api is now gone.

src/components/Block/Block.styles.ts
```diff
@@ -15,19 +17,38 @@ const getMarginTop = (topSpacing?: GutterSize, push?: number) => {
...
+  [TextSize.LARGE]: '1.5rem',
+  [TextSize.MEDIUM]: '1.4rem',
+  [TextSize.MEDIUM_SUB]: '1.4rem',
+  [TextSize.SMALL]: '1.2rem',
+  [TextSize.XSMALL]: '1.0rem',
+};
+
 export const getStyles = (props: BlockProps): IRawStyle => {
   const { push, textSize, topSpacing, bottomSpacing, textAlign, textColor } = props;
+  const theme = getTheme();
```
chrismohr: Todo in another PR: add memoization

src/components/Text/Text.styles.ts
```diff
@@ -1,38 +1,55 @@
...
+  [TextSize.MEDIUM_SUB]: '1.4rem',
+  [TextSize.SMALL]: '1.2rem',
+  [TextSize.XSMALL]: '1.0rem',
 };
 
-export const getStyles = (props: TextProps): IRawStyle => {
-  const { size, maxWidth, bold, uppercase, color } = props;
+const getMemoizedClassNames = memoizeFunction((styleProps: TextProps, theme) => {
+  const { size, maxWidth, bold, uppercase, color } = styleProps;
+  const font = size ? theme.fonts[size === TextSize.MEDIUM_SUB ? 'smallPlus' : size] : undefined;
```
chrismohr: Todo in a future PR: rename MEDIUM_SUB to SMALL_PLUS to align with Fabric.
https://github.com/Microsoft/YamUI/issues/424

src/components/Block/Block.styles.ts
unindented: Should we try adding a `Customizer` in one of our consumer apps? I'm kind of excited to see this working.

chrismohr: Yeah, me too. I'll look at this next.

src/components/ChoiceGroup/ChoiceGroup.styles.tsx
```diff
@@ -1,17 +1,18 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
-import { palette } from '../../util/colors';
+import { getTheme } from '../../util/colors';
 import { IChoiceGroupStyles } from 'office-ui-fabric-react/lib/components/ChoiceGroup/ChoiceGroup.types';
 import {
   IChoiceGroupOptionStyles,
   IChoiceGroupOptionStyleProps,
 } from 'office-ui-fabric-react/lib/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.types';
-import { textColors, fontSizes, fontWeights } from '../../util/styles/fonts';
+import { textColors, fontWeights } from '../../util/styles/fonts';
```
unindented: Why do we still get these things from our files instead of the theme?

unindented: Ah, this is `ChoiceGroup`, so not directly related to this PR.

PR #404, opened 2018-06-13
New component: ModalDialog.

swese44:
A ModalDialog is used as a temporary modal UI overlay to provide
contextual app information or ask the user for input.

This component wraps the Modal component from office-ui-fabric-react.

TODO:
- [x] Visual diffs
- [ ] Accessibility review
- [ ] Remove Clickable references in .md file, update example selectors to capture visual diffs on open dialog component

## Pull request checklist


swese44:
@BrendanMcK Fabric's Modal only closes on ESC when a tabbable item within the modal content is selected. if this is a problem we should probably drive the solution separately in Fabric directly

swese44:
(@mloreti will be taking over this PR, @smoak had done the initial work)

swese44:
Modal is the utility component to render arbitrary content in a `<Popup>`, chich manages the underlying functionality like ESC and click-out to close. Scott looked at Dialog first but it would require some upstream changes in Fabric:
1. title accepts a string, not JSX, so if we want it in an `<h1>` tag we'd have to contribute an `onRenderTitle` prop
2. we also need to render our own close Icon, so we'd have to contribute an `onRenderCloseButton` prop

Scott had opted instead for an `onRenderHeader` prop where we'd render them both, but it seemed like the actual Dialog component wasn't actually doing much for us at that point. Since Modal's example shows a custom rendered header we figured this would be the best approach.

@mloreti is going to take over this PR for @smoak, maybe you two could pair up and decide if this is the right approach or if we should make the changes to Fabric's Dialog component?

swese44:
@BrendanMcK `Fabric's Modal only closes on ESC when a tabbable item within the modal content is selected. if this is a problem we should probably drive the solution separately in Fabric directly`

both Fabric's Modal and Dialog have this behavior, i assume it's the behavior of the underlying `<Popup>` component and if it's an issue we should separately drive a fix there

BrendanMcK:
* so if we want it in an <h1> tag 

Turns out that Fabric's Dialog does use role="heading", which is *almost* equivalent - they might need to add an aria-level=1. But IMO fix here is to get Fabric to change.

* we also need to render our own close Icon

Does ours have different visuals than theirs?

* only closes on ESC when a tabbable item within the modal content / both Fabric's Modal and Dialog have this behavior

That does sound like a Fabric issue; it repros with their Modal example. Curious, we we know of any cases where this would happen IRL? Generally a dialog will have some control in it that has focus - often the close button as a last resort if there's no other appropriate control. So maybe this isn't an issue so long as there's well-formed content?

* seemed like the actual Dialog component wasn't actually doing much for us

Ya, so it's sounding like it's just adding some layout/heading policy that we don't need.

TL;DR: no objections from me on Modal vs Dialog, so long as the end result works well; just want to understand why we're not using the Fabric component that would appear at first glance to be the closest match.

swese44:
up to this point it's been a hard requirement from Design that we use the YamUI Icons and not any Fabric icons. please follow the PR along next week and provide any necessary feedback, thanks!

mloreti:
@swese44 @unindented @richardtagger @chrismohr @BrendanMcK 
Added the escape key listener. Can you take a look?

src/components/ModalDialog/ModalDialog.tsx
```diff
@@ -46,22 +47,34 @@ export interface ModalDialogProps extends BaseComponentProps {
...
+      isOpen: props.isOpen,
+    };
+  }
+
   public render() {
-    const { children, className, isOpen, onDismissed, size } = this.props;
+    const { children, className, onDismissed, size } = this.props;
     const classNames = getClassNames(size);
 
+    this.state.isOpen ? this.startKeyListener() : this.stopKeyListener();
```
richardtagger: Don't add listeners as part of render. Perhaps componentDidUpdate.

src/components/ModalDialog/ModalDialog.tsx
```diff
@@ -98,4 +111,20 @@ export default class ModalDialog extends React.Component<ModalDialogProps> {
...
+      }));
+    }
+  };
+
+  private startKeyListener = () => {
+    document.addEventListener('keydown', this.handleKeyDown);
+  };
+
+  private stopKeyListener = () => {
+    document.removeEventListener('keydown', this.handleKeyDown);
```
richardtagger: this needs to happen on componentDidUnmount too right?

BrendanMcK:
Per @swese44 above, turns out we don't need to add our own ESC handler here - Fabric's Modal is already doing that for us via the child Popup. Here's a summary of what's going on within Fabric:
* Dialog adds some basic layout policy (title bar, etc) on top of Modal
* Modal creates a Popup in a Layer, and also adds an Overlay
* Popup handles ESC and restoring focus.

Since we're deriving from Modal, we only need to duplicate the specific items that Dialog is doing - setting up the title bar. Things to keep in mind from a11y point of view here are:
* Having a close button with appropriate alternate text
* Having a H1 or equivalent (eg. DIV with role=heading) as the title bar
* Having aria-labelledby on the role=dialog element refer to the H1 by ID. Looks like this can be set as a prop to the Modal.

src/components/ModalDialog/ModalDialog.tsx
mloreti: Yup, makes sense.

src/components/ModalDialog/ModalDialog.tsx
mloreti: Got it.

mloreti:
@BrendanMcK Ahh, didn't catch that. So ESC only works when a tab-able item is selected which is odd to me. I don't see a conclusion if that's the desired behavior. Should we consult with design?

BrendanMcK:
> So ESC only works when a tab-able item is selected which is odd to me.

I see this as a defect in Fabric's Popup/Modal stuffs. I think what's happening here is that:
- Popup is only listening to ESC keypresses coming from within its subtree.
- Modal has not found a default item within itself to set focus to
So focus is remaining outside the popup, and ESC is getting ignored.

I don't think this should be an issue for us, though, since all of our dialogs should have *some* focusable item - even if that's just the close button.

Either way, I think we're safe treating this as a Fabric issue.

I'm spelunking in other Fabric lightbox stuffs at the moment, and have [opened a couple of issues](https://github.com/OfficeDev/office-ui-fabric-react/issues/created_by/BrendanMcK) against them - I'll add this one in too.

mloreti:
@BrendanMcK Got it. I'll remove it then.

src/components/ModalDialog/ModalDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+  closeAriaLabel: string;
+
+  /**
+   * The size of the ModalDialog
+   */
+  size: ModalDialogSize;
+}
+
+export interface ModalDialogState {
+  isOpen: boolean;
```
unindented: Why have `isOpen` as both a prop and a state key? Shouldn't we decide whether this is a controlled or uncontrolled component? I think all other stateful components in YamUI are controlled (e.g. `Dropdown`).

mloreti: Yeah I realized that after discussing with Arturo. I removed state completely and it worked fine but now I'm going to remove everything that I've added as it's no longer necessary.

BrendanMcK:
Another thing to consider - we'll need to make sure the 'Close' button label is localized, so our component will need to take a closeButtonAriaLabel prop or similar - that's the same name that Fabric use for their prop, btw.

smoak:
@BrendanMcK The following a11y bits are in here now:

- [x] Having a close button with appropriate alternate text (I believe this is achieved by using the `Clickable` component)
- [x] Having a H1 or equivalent (eg. DIV with role=heading) as the title bar (This is achieved by using the `Heading` component)
- [x] Having aria-labelledby on the role=dialog element refer to the H1 by ID. Looks like this can be set as a prop to the Modal. (This is achieved by wrapping the header with a div with id y-modalDialog--title and titleAriaId="y-modalDialog--title" on the FabricModal component)

Can you double check we aren't missing anything?

BrendanMcK:
Looks good from a11y point of view.

Curious about hideTitle use case - it also hides the close button; should it do that? What are the use cases where we'd use this?

One that comes to mind is the terms of use - but in that case, we want to hide the close button, not the title. Or is the idea that an instance can use this to hide the entire header to supply their own (eg. if they want a search box in their own header area)?

BrendanMcK:
Something doesn't feel quite right about the close handling logic here; partly due to Fabric's underlying logic being a bit unclear too.

Generalizing, what we'd want is:

- some event/callback that indicates a *request to close*. This means ESC has been pressed, user has clicked the background, or clicked the close button. The owner of a dialog will typically respond by actually closing the dialog, but might reject this - for example, our compose lightbox will ask 'are you sure?' if there's body text or attachments

- some event/callback that indicates that the lightbox has actually closed.

We're currently treating clicking the close button the same as onDismiss and onDismissed, when they're really not the same thing.

BrendanMcK:
I think the thing to do here is figure out if Fabric's onDismiss is actually a rejectable reqest-to-dismiss, in which case it can be treated the same way as the close button; while the onDismissed a separate lightbox-has-been-closed event. May want to see what Fabric's own Modal is doing with these.

smoak:
@swese44 I don't understand what you mean by:

`Remove Clickable references in .md file, update example selectors to capture visual diffs on open dialog component`. Is there an example component that does this already?

BrendanMcK:
Re `update example selectors to capture visual diffs on open dialog component` - issue here is that by default, the snapshot stuffs would just take a snapshot of the trigger button, not the open lightbox. So seems we need to work with the props [here](https://github.com/Microsoft/YamUI/pull/404/commits/b9215af53f667f069f6953b4f60fb8dc066d03ff#diff-f53b604a00d9e8e2eb7726e473e9d677R9) to tell the snapshot code how to invoke the dialog so it can capture it.

Tooltip.md is doing something a bit similar, using props to get the snapshot tool to hover over the target to get the tooltip to appear so it can capture it.

smoak:
@BrendanMcK I think the issue is that the lightbox is out of view of the screenshot tool, so you only see the trigger button with the grey overlay

BrendanMcK:
Ya, that's going to be an issue too. @unindented , any thoughts? If we could tell the snapshot to use a different selector to capture instead of the original block, that might work here.

Turns out that Fabric's Modal puts role=dialog on one of the outermost elements that contains the grey overlay shade; so that element takes up the *entire* screen content. The portion that we think of as the actual dialog itself, and which we'd like to snapshot - ie the content - is *within* that. So we can't use say [role=dialog] as a selector here. But we could put a classname on our own dialog contents, and look for that.

smoak:
> If we could tell the snapshot to use a different selector to capture instead of the original block, that might work here.

Talked with @unindented, going with your suggestion @BrendanMcK

smoak:
PR with an added feature to enable it: https://github.com/unindented/react-styleguidist-visual/pull/33

smoak:
Not really sure what's meant by "Remove Clickable references in .md file, update example selectors to capture visual diffs on open dialog component", but there are visual diffs now. @BrendanMcK anything else for accessibility? I think this should hopefully be good to go now.

smoak:
@BrendanMcK I have added your suggestion. LGTU?

BrendanMcK:
LGTM! - I don't have approval rights for this repo, though, so you'll still need one of chrismohr, richardtagger, and/or unindented to sign off.

smoak:
@unindented LGTU?

unindented:
What's this screenshot about?

screen shot 2018-06-22 at 15 19 59

smoak:
Not sure, it got generated during `npm run test:visual`

smoak:
Oh looking more closely, I defined a second exampe for a hidden title but didnt update it to use the new preview selector

smoak:
@unindented Updated

PR #405, opened 2018-06-13
Support multiple item templates in SuggestionsList

### LINKED ISSUE #167: SuggestionsList missing template for results other than User

swese44:
- Implements item templates to support users, groups, links, etc. Fixes #167
- Extracts `SuggestionsListItemContent.tsx` so it can be reused in pickers

## Pull request checklist


src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -68,16 +70,16 @@ export default class SuggestionsList extends React.PureComponent<SuggestionsList
       <div className={join(['y-suggestions-list', 'y-hc-border', className, classNames.root])}>
         {hasResults && (
           <LayoutList direction="vertical">
-            {groupedItems.map((group: SuggestionItemGroupProps) => (
+            {groupedItems.map((group: SuggestionItemGroup) => (
```
richardtagger: this is a fairly miserable mount of nesting - can this be pulled out into a this.getGroupedItem or whatever?

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -68,16 +70,16 @@ export default class SuggestionsList extends React.PureComponent<SuggestionsList
...
               <LayoutListItem key={group.title}>
                 <LayoutList direction="vertical">
                   {group.items.map((item: SuggestionItem) => (
```
richardtagger: same here-  i know you didnt add this but the can this be pulled out into a this.getItem or whatever?

src/components/SuggestionsList/SuggestionsList.tsx
```diff
@@ -68,16 +70,16 @@ export default class SuggestionsList extends React.PureComponent<SuggestionsList
       <div className={join(['y-suggestions-list', 'y-hc-border', className, classNames.root])}>
         {hasResults && (
```
richardtagger: never a fan of inline logic in JSX - cant tell from this snippet - should the component not just render `null` if !hasResults, or if not something like `const results = hasResults ? this.getResults() : null;`

src/components/SuggestionsList/SuggestionsListItem.test.tsx
```diff
@@ -29,42 +31,21 @@ describe('<SuggestionsListItem />', () => {
...
-    it('renders as expected', () => {
-      expect(rendered).toMatchSnapshot();
-    });
-  });
-
-  describe('with avatarProps', () => {
-    beforeEach(() => {
-      const avatarProps = { borderType: BorderType.SOFT };
-      rendered = shallow(<SuggestionsListItem {...getProps({ avatarProps })} />);
+      component = shallow(<SuggestionsListItem {...baseProps({ isSelected: true })} item={textItem} />);
```
richardtagger: Why not supply textItem to `baseProps` (weird naming IMO btw).

```ts
const props = getProps({ isSelected: true, item: textItem });
component = shallow(<SuggestionsListItem {...props} />);
```

src/components/SuggestionsList/SuggestionsListItem.test.tsx
```diff
@@ -29,42 +31,21 @@ describe('<SuggestionsListItem />', () => {
...
+      expect(component).toMatchSnapshot();
     });
   });
 
   describe('when selecting', () => {
     let preventDefault: jest.Mock;
 
     beforeEach(() => {
-      rendered = shallow(<SuggestionsListItem {...getProps()} />);
+      component = shallow(<SuggestionsListItem {...baseProps()} item={textItem} />);
```
richardtagger: same as above

src/components/SuggestionsList/SuggestionsListItemContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  private getImage(item: MediaSuggestionItem) {
+    if (!item.imageUrl) {
+      return null;
+    }
+
+    const borderType = item.template === SuggestionsListItemTemplate.MEDIA_SOFT ? BorderType.SOFT : undefined;
+    return (
+      <span aria-hidden={true}>
+        <Image
```
richardtagger: just musing - but should we allow aria-hidden on all YamUI components so we dont need to create extra DOM nodes like this

src/components/SuggestionsList/SuggestionsList.tsx
swese44: i agree, i'll update these

src/components/SuggestionsList/SuggestionsList.tsx
swese44: 👍

src/components/SuggestionsList/SuggestionsList.tsx
swese44: ya i'll look at these

src/components/SuggestionsList/SuggestionsListItem.test.tsx
swese44: ya things changed when i moved the item from splat props to its own prop. i'll update

src/components/SuggestionsList/SuggestionsListItemContent.tsx
swese44: ya not a bad idea, we can add an issue to track. image should probably support it directly for cases like this

src/components/SuggestionsList/SuggestionsList.tsx
swese44: done

src/components/SuggestionsList/SuggestionsList.tsx
swese44: done

src/components/SuggestionsList/SuggestionsList.tsx
swese44: done

src/components/SuggestionsList/SuggestionsListItem.test.tsx
swese44: done

src/components/SuggestionsList/SuggestionsListItem.test.tsx
swese44: done

src/components/SuggestionsList/SuggestionsList.test.tsx
```diff
@@ -59,51 +60,50 @@ describe('<SuggestionsList />', () => {
...
 
       describe('when onHover is called', () => {
         beforeEach(() => {
-          rendered.find('SuggestionsListItem').simulate('hover', '1');
+          component.find('SuggestionsListItem').simulate('hover', '1');
         });
 
         it('updates the state', () => {
-          expect(rendered.state()).toMatchSnapshot();
+          expect(component.state()).toMatchSnapshot();
```
richardtagger: I know you didn't do this but on reflection snapshotting state is mostly a smell to me - we should care about side effects (in this case the rendered output). The internal state representation is an implementation detail. I think it's fine to keep it but just thinking out loud

swese44: ah i saw this too and meant to circle back to it. i don't know why this test is here because this component doesn't even track any state. i will remove it

PR #408, opened 2018-06-14
feat(utilities): import getFirstFocusable method from Fabric

eoh:
Importing `getFirstFocusable` method from Fabric, to be used with `SkipLinks` component in MC


src/util/getFirstFocusable
```diff
@@ -0,0 +1,4 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import { getFirstFocusable } from '@uifabric/utilities/lib/focus';
+
+export { getFirstFocusable };
```
unindented: This can be condensed into one line:

```ts
export { getFirstFocusable } from '@uifabric/utilities/lib/focus';
```

unindented:
Oh shit, I didn't notice. That's what we get for not writing tests. @eoh can you create a new PR that adds the extension and a unit test?

BrendanMcK:
Is there a meaningful test that can be written here, though? Or is it the case that the import itself *is* the test, with an empty describe? Because we're exporting directly from fabric, it's not like we even have an intermediate wrapper class to exercise.

unindented:
Maybe just importing it and expecting that the thing you got is a function?

PR #412, opened 2018-06-16
Initial Picker implementation

swese44:
This is a single Picker component which renders items of type SuggestionItem (which already exists from SuggestionsList and has templates for each of our use cases). The application would be responsible for passing items of any `SuggestionItem` type and the YamUI Picker just renders them. i.e. the app can manage a UserPicker, GroupPicker, TopicPicker, CCPickerBar (1 Group + Many Users). For a case where one group and many users could be selected, the application would be responsible for showing the selected group first in the array of selectedItems, and not passing any more group items to the list of suggestions.

screen shot 2018-06-15 at 5 55 56 pm

screen shot 2018-06-15 at 5 56 09 pm

screen shot 2018-06-15 at 5 56 16 pm

screen shot 2018-06-15 at 5 56 21 pm

screen shot 2018-06-15 at 5 56 41 pm

screen shot 2018-06-15 at 5 56 51 pm

TODO:
- [ ] Need a Designer to signoff on the selected item pills, might need full mockups i just scrapped this together. we probably want a different treatment for tags vs. groups vs. people.
- [ ] Need accessibility review. The selectedItem pills were based on TagPicker's implementation but we can render them however we want.
- [ ] When disabled (triggered by reaching the item limit) overall height is like 2px shorter for some reason

## Pull request checklist


src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,156 @@
...
+        onResolveSuggestions={onResolveSuggestions}
+        pickerSuggestionsProps={{
+          suggestionsHeaderText,
+          noResultsFoundText,
+        }}
+        itemLimit={itemLimit}
+        disabled={disabled}
+        inputProps={{
+          onBlur: onBlur,
+          onFocus: onFocus,
```
richardtagger: object shorthand?

src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    return <SuggestionItemContent item={item} className={classNames.suggestionItem} />;
+  };
+
+  private getSelectedItem = (props: IPickerItemProps<SuggestionItem>) => {
+    const classNames = getClassNames();
+
+    return (
+      <div
+        role={'listitem'}
```
richardtagger: is there a reason to use div with role=listitem rather than an `<LI>`? is it because it might not be in a UL or OL

src/components/Picker/Picker.tsx
richardtagger: I guess we dont lint for it in YamUI?

unindented: Yeah, difficult to keep the two sets of configs in sync. Will fix.

src/components/Picker/Picker.tsx
unindented: I think @swese44 may have taken this from Fabric's `TagPicker` implementation. I'm not sure why they're doing this.

PR #413, opened 2018-06-16
Remove Context API - Use 'selectors' styles instead

### LINKED ISSUE #411: Should we replace all Context API usage with selectors in our getClassNames() functions?
https://www.yammer.com/microsoft.com/#/Threads/show?threadId=1107212682

Related, note that Block's 100% width rule should only apply when it is a direct descendant of Clickable.

chrismohr:
The use of the context api here was causing test failures in downstream dependencies, since it is not yet fully supported by enzyme. 

We were using context to achieve styles that were dependent on a specific nesting of components. 

Instead, we can use the `selectors` facility offered by merge-styles to achieve is, which I think is an overall better solution anyway.

fixes #411

### DIFF
--- a/src/components/Clickable/Clickable.styles.ts
+++ b/src/components/Clickable/Clickable.styles.ts
@@ -24,7 +24,7 @@ export const getClassNames = memoizeFunction((styleProps: ClickableStyleProps) =
       display: block ? 'block' : 'inline',
       width: block ? '100%' : 'undefined',
       selectors: {
-        '.y-block': {
+        '> .y-block': {
           width: '100%',
         },
       },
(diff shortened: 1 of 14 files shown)


src/components/Clickable/Clickable.styles.ts
```diff
@@ -23,6 +23,11 @@ export const getClassNames = memoizeFunction((styleProps: ClickableStyleProps) =
       cursor: 'pointer',
       display: block ? 'block' : 'inline',
       width: block ? '100%' : 'undefined',
+      selectors: {
+        '.y-block': {
```
swese44: this should be a direct descendant selector, i can update

PR #416, opened 2018-06-19
refactor(utilities): added .ts extension to getFirstFocusable, added unit test

eoh:


src/util/getFirstFocusable.test.ts
```diff
@@ -0,0 +1,12 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+jest.mock('@uifabric/utilities/lib/focus', () => ({
+  getFirstFocusable: jest.fn(),
+}));
```
unindented: Why mock it? If you replace it with a spy, and then check that it's a function, it's always going to return true! I could go into the implementation file, replace its contents with `export const getFirstFocusable = 42` and the tests would still pass.

src/util/getFirstFocusable.test.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { getFirstFocusable } from '@uifabric/utilities/lib/focus';
+
+describe('getFirstFocusable', () => {
+  it('checks if import is a function', () => {
+    expect(typeof getFirstFocusable).toBe('function');
```
unindented: Another way of doing this would be:

```
expect(getFirstFocusable).toEqual(expect.any(Function));
```

I'm fine with either.

src/util/getFirstFocusable.test.ts
eoh: Good point.  I get a 'Jest encountered an unexpected token' error without mocking, the linter complains about an incorrect header in `@uifabric/utilities/lib/focus.js`

Will adding `transformIgnorePatterns` to YamUI's jest config fix this?

config/jest/config.json
```diff
@@ -12,6 +12,7 @@
     "<rootDir>/src/demos"
   ],
   "moduleNameMapper": {
+    "@uifabric/utilities/(.*)": "<rootDir>/node_modules/@uifabric/utilities/",
```
unindented: Don't understand why this is working, cause you're mapping a module to the same module...

eoh:
@unindented would you mind merging, i don't have the power :(

PR #420, opened 2018-06-26
8.5

eoh:
Releasing a new version to get ```getFirstFocusable``` into MC


unindented:
@smoak already released version 8.5.0, it's on npm.

PR #426, opened 2018-06-27
add customizer component

chrismohr:
- allow Button component to have a custom theme
- add Customizer component that accepts a theme

## Background:
Fabric offers two ways to set a theme. 

1. loadTheme - https://github.com/OfficeDev/office-ui-fabric-react/blob/f98fd419be0c40c5c22d70bcd60f114dc445c4fa/packages/office-ui-fabric-react/src/components/Theme/docs/ThemesOverview.md#overriding-themes

2. Customizer - partially documented here: https://github.com/OfficeDev/office-ui-fabric-react/blob/f98fd419be0c40c5c22d70bcd60f114dc445c4fa/packages/variants/README.md

- components that need to consume theme, will need to be decorated with `@customizable` and accept `theme` as a prop

We prefer Customizer because it will allow us to sidestep any issues that result from setting the theme globally with loadTheme. (potential pollution if multiple instances of our app are on the same page etc.)

## Pull request checklist

* [x] Fix coverage for Button.styles

## Next step
Decorate components that need to consume theme with `@customizable` and have props implement `CustomizableComponentProps`.


src/components/Customizer/Customizer.md
```diff
@@ -0,0 +1,24 @@
+### Examples
+
+Button With Theme:
```
unindented: Why All Capitals?

chrismohr: haha... fixed

src/components/Customizer/Customizer.tsx
```diff
@@ -1,26 +1,19 @@
...
  * Used to inject a custom theme object into customizable components, overriding the global theme.
  */
 export default class Customizer extends React.Component<CustomizerProps> {
   public render() {
-    const { theme } = this.props;
-
-    return <FabricCustomizer settings={{ theme }}>{this.props.children}</FabricCustomizer>;
+    return <FabricCustomizer {...this.props} />;
   }
 }
```
unindented: Why not `export default FabricCustomizer` directly?

chrismohr: @unindented I tried, but that removed it from the style guide.  

There are some related docs here, but it wasn't clear to me how else to force it back in.

chrismohr: https://react-styleguidist.js.org/docs/thirdparties.html

PR #427, opened 2018-06-28
make AsideSection customizable

chrismohr:
A step towards making all components that use theme to be `customizable`.

In the next PR, I hope to do several/all components, just want to get eyes on this to make sure I'm setting up a good pattern.

### DIFF
--- a/src/components/AsideSection/AsideSection.test.tsx
+++ b/src/components/AsideSection/AsideSection.test.tsx
@@ -1,15 +1,16 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 import * as React from 'react';
-import { shallow, ShallowWrapper } from 'enzyme';
-import Clickable from '../Clickable';
+import { shallow, mount, ShallowWrapper } from 'enzyme';
 import AsideSection, { AsideSectionProps } from '.';
+import Clickable from '../Clickable';
+import Customizer from '../Customizer';
 
 describe('<AsideSection />', () => {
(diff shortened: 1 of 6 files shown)


src/components/Button/Button.styles.ts
```diff
@@ -99,9 +98,8 @@ const getMemoizedBaseButtonStyles = memoizeFunction((props, theme): IButtonStyle
   };
 });
 
-export const getBaseButtonStyles = (props: BaseButtonStyleProps) => {
-  const theme = props.theme || getTheme();
```
chrismohr: Good news is that this logic is built into `customizable`.  If were not wrapped in a `Customizer`, it will pull the global theme.

unindented: Nice.

src/components/AsideSection/AsideSection.test.tsx
```diff
@@ -34,11 +35,23 @@ describe('<AsideSection />', () => {
...
+    it('receives custom theme', () => {
+      const theme = { palette: {}, semanticColors: {} };
+      const mountedComponent = mount(
+        <Customizer settings={{ theme }}>
+          <AsideSection title="TITLE" />
+        </Customizer>,
+      );
+      expect(mountedComponent.find('AsideSection').prop('theme')).toBe(theme);
+    });
+  });
```
chrismohr: I'm planning to repeat this test for every `customizable` component.

However it's not yet entirely clear to me why I wasn't able to pull the theme prop using a `shallow` render with `dive`. Is `mount` fine here?

PR #428, opened 2018-06-28
Customizable Avatar

chrismohr:
- now exporting an uncustomizable component to help with testing (no more need to `dive()`), and to get around a challenge of getting test coverage in the case where a theme is absent


src/components/AsideSection/AsideSection.test.tsx
```diff
@@ -45,13 +45,13 @@ describe('<AsideSection />', () => {
 
   describe('with customizer', () => {
     it('receives custom theme', () => {
-      const theme = { palette: {}, semanticColors: {} };
+      const theme = nullTheme;
```
chrismohr: maybe `defaultTheme` would be a better name.

src/components/Customizer/Customizer.tsx
```diff
@@ -1,18 +1,25 @@
...
 
-export { createTheme, ITheme, CustomizerProps };
+export { createTheme, ITheme, ICustomizerProps };
+
+export interface CustomizableComponentProps extends BaseComponentProps {
+  /**
+   * Theme used to style the component. This will often be provided by `customizable`.
+   */
+  theme?: ITheme;
+}
```
chrismohr: I moved this from `src/util/BaseComponent/props.ts` because I wanted it to be in the same place as the `nullTheme`, and `nullTheme` didn't seem to fit well in `props.ts`.

src/components/AsideSection/AsideSection.tsx
```diff
@@ -24,10 +25,9 @@ export interface AsideSectionProps extends NestableBaseComponentProps {
 /**
  * A section component to be used primarily for sidebar modules
  */
-@customizable('AsideSection', ['theme'])
-export default class AsideSection extends React.Component<AsideSectionProps & CustomizableComponentProps> {
+export class AsideSection extends React.Component<AsideSectionProps & CustomizableComponentProps> {
```
chrismohr: This breaks the pattern a bit, since `AsideSectionProps` extends `NestableBaseComponentProps` instead of `BaseComponent` props.

Not sure if there are tradeoffs between that and using intersection like this.

src/components/AsideSection/AsideSection.types.ts
```diff
@@ -0,0 +1,14 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
```
chrismohr: I chose to follow the fabric file naming convention here.

src/components/Customizer/Customizer.tsx
chrismohr: *defaultTheme

chrismohr:
@unindented and I talked about various tradeoffs here. 

We considered casting theme to be non-null, excluding the theme assignment line of code from coverage reporting, lowering the coverage threshold, but liked the tradeoffs of this solution the best.

src/components/AsideSection/AsideSection.test.tsx
```diff
@@ -45,13 +46,13 @@ describe('<AsideSection />', () => {
...
   describe('with customizer', () => {
     it('receives custom theme', () => {
-      const theme = { palette: {}, semanticColors: {} };
+      const theme = defaultTheme;
       const mountedComponent = mount(
         <Customizer settings={{ theme }}>
-          <AsideSection title="TITLE" />
+          <CustomizableAsideSection title="TITLE" />
         </Customizer>,
       );
```
unindented: Would it be better to move these lines above to a `beforeEach` associated to the `with customizer` block?

chrismohr: Yeah, that would be more consistent with other tests. Done.

PR #430, opened 2018-06-29
derive icon sizes in block from theme

chrismohr:
Previously the icon size in a block that had TextSize of `medium` or `large` was hard-coded to be smaller than the text. Is there a good reason to keep it like this?

Since we're now getting the font sizes for text from the theme, we also need a way to get the icon sizes to not be hard-coded.  Accepting this visual change is the most straight forward way I can think of.


swese44:
Design spent a lot of time working out these specific icon sizes at each font size, please reach out to Aaron before merging

chrismohr:
@swese44 For sure.  I pinged Aaron in the UI domain group when I opened this, and I'll wait for his input.

swese44:
from my end LGTM, this change makes a lot of sense

swese44:
note that `<Text>` needs the exact same nested icon logic, you should probably update it as part of this PR

chrismohr:
@swese44  Good point.  I'll do that now.

swese44:
the larger icon sizes might push the line-heights taller in the non-block-level `<Text>` components and break our vertical rhythm, this may have been the primary reason why those specific font sizes use smaller icons. if so, we'll have to keep some sort of mapping to use smaller icons in those cases

swese44:
nevermind that, diffs look good

swese44:
actually the visual diff is only using `<Block>`. i think we should duplicate that markdown example replacing `<Block>` with `<Text>` so we can make sure the nested icons don't affect line-heights

swese44:
(sorry to be nitpicky, i just remember icons + font sizes + line heights to be really tricky to get right)

chrismohr:
Thank you for being nitpicky.  I don't have as much context on why it is the way it is, so it's really helpful.  I'll add a Text example.

chrismohr:
@swese44 I've added the example, and from what I can tell, the line height is fine.  I'm looking at the `computed` height value in the css inspector.  For medium it's 40px, and large is 48px. (for 2 lines)  Am I looking at the right thing?

lineheight

swese44:
each line should have overall height equal to that size's line-height. if medium and mediumSub have line-height 20px, one line should be 20px tall and 2 lines wrapped should be 40px tall. if these all match up then we're good

chrismohr:
Cool.  I just double-checked, and looks like we're good in both cases.  Now I'll just wait for @vanschyndel 's approval.

PR #443, opened 2018-07-02
feat(container query): add size observer component

jessemartin:
This introduces a SizeObserver component, which allows media-query like rendering of a component
based on its height and width.

## Pull request checklist


src/components/ContainerQuery/ContainerQuery.tsx
```diff
@@ -0,0 +1,126 @@
...
+  target: HTMLElement;
+}
+
+type ResizeObserver = new (cb: (entries: ResizeObserverEntry[]) => void) => {
+  disconnect(): void;
+  observe(): void;
+  unobserve(): void;
+};
+
+export interface ContainerQueryProps extends NestableBaseComponentProps, FocusableComponentProps {
```
swese44: why FocusableComponentProps here?

src/components/ContainerQuery/ContainerQuery.test.tsx
```diff
@@ -0,0 +1,47 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import * as React from 'react';
+import ContainerQuery from '.';
+import { create as createRenderer, ReactTestRendererJSON } from 'react-test-renderer';
```
swese44: Why not use enzyme like the other test files? Just curious

src/components/ContainerQuery/ContainerQuery.tsx
```diff
@@ -0,0 +1,126 @@
...
+  public componentDidMount() {
+    this.observeElement();
+  }
+
+  public componentWillUnmount() {
+    this.resizeObserver.disconnect();
+  }
+
+  public render() {
+    return <div ref={this.observedElement}>{this.getContent()}</div>;
```
swese44: What happens when wide child content is displayed at a wider query and then the viewport is resized smaller than the content can shrink -- if the child content is too wide will we not be able to responsively switch to a smaller query? I wonder if the render method needs to look like this? Might also need some positioning magic so the child wrapper div doesn't push the wrapper too wide?

```
<div>
    <div ref={ref} />
    <div>
        {this.getContent()}
    </div>
</div>
```

unindented: Shouldn't we ensure that children are flexible instead?

src/components/ContainerQuery/ContainerQuery.tsx
unindented: I think it was copy-pasta. I'll fix.

src/components/ContainerQuery/ContainerQuery.test.tsx
unindented: We also seem to be doing this for `Clickable`, not sure why.

src/components/ContainerQuery/ContainerQuery.tsx
jessemartin: Yes, it was. Let's update.

src/components/ContainerQuery/ContainerQuery.test.tsx
jessemartin: Yeah there's no need to be doing this, we should move to enzyme. (This was copied from the clickable test)

jessemartin:
@unindented `SizeObserver` or `ResizeObserver` make sense to me.

src/components/SizeObserver/SizeObserver.tsx
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if (maxWidth !== undefined && width > maxWidth) {
+      continue;
+    }
+    if (minWidth !== undefined && width < minWidth) {
+      continue;
+    }
+
+    return key;
+  }
```
chrismohr: It took me a little while to grok this with the `continue`s, how would you feel about this refactor?
```
const getMatchingQuerySize = (query: SizeQueriesMap, height: number, width: number) => {
  for (const key of Object.keys(query)) {
    const {
      maxHeight = Number.MAX_SAFE_INTEGER,
      maxWidth = Number.MAX_SAFE_INTEGER,
      minHeight = 0,
      minWidth = 0,
    } = query[key];
    if (height > maxHeight && height < minHeight && width > maxWidth && width < minWidth) {
      return key;
    }
  }
};
```

jessemartin: I agree this is more understandable, thanks! We will update.

src/components/SizeObserver/SizeObserver.md
```diff
@@ -0,0 +1,29 @@
+### Notes for use
+
+SizeObserver allows for conditional rendering based on the component's height or width.
+
+
+### Examples
+
+```js { "props": { "query": { "small": { "maxWidth": 500 }, "large": { "minWidth": 501 } } } }
```
chrismohr: For what reason are these props are here?

Could it not just be ```js ?

src/components/ChoiceGroup/ChoiceGroup.test.tsx
```diff
@@ -7,7 +7,7 @@ import { IChoiceGroupOptionProps } from 'office-ui-fabric-react/lib/components/C
 
 describe('<ChoiceGroup />', () => {
   let component: ShallowWrapper<ChoiceGroupProps>;
-  let labelWithSublabel: JSX.Element;
```
chrismohr: Mostly for my learning, what caused you to replace `JSX.Element`s?

src/components/SizeObserver/SizeObserver.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+        case 'large':
+          return <span>Large</span>;
+        case 'small':
+          return <span>Small</span>;
+        default:
+          return <span>Other</span>;
+      }
+    });
+
+    component = mount(<SizeObserver query={query} renderAt={renderAt} />);
```
chrismohr: Looks to me like you'd get the same coverage, with clearer snapshots by just looking at the `span`.

```    
component = mount(<SizeObserver query={query} renderAt={renderAt} />).find('span');
```

chrismohr: Or maybe `find('div')` to get the wapping div too?

src/components/ChoiceGroup/ChoiceGroup.test.tsx
jessemartin: @chrismohr 
JSX.Element is an alias to React.ReactElement.
However, React.ReactElement only describes react classes and stateless functional components. It does not include support for `Fragment` or `Portal`.
React.ReactNode extends React.ReactElement and includes support for Fragments and Portals.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L151

I think @unindented made these changes so our types are explicit and support Fragments.

chrismohr: Cool.  Thanks for the explanation.

src/components/SizeObserver/SizeObserver.test.tsx
jessemartin: Agreed. I will update.

src/components/SizeObserver/SizeObserver.md
jessemartin: I see we declare the props like this in all of the component readme's I've looked at. I'm not sure why, though. @unindented  @swese44 @richardtagger are these used for anything?

src/components/SizeObserver/__snapshots__/SizeObserver.test.tsx.snap
```diff
@@ -0,0 +1,19 @@
...
+
+exports[`<SizeObserver /> renders at the default size on first render 1`] = `
+<span>
+  Other
+</span>
+`;
+
+exports[`<SizeObserver /> when its size changes renders at the corresponding size 1`] = `
+<span>
+  Other
```
chrismohr: Is this the expected result?  Looking at the test, I would have expected `small`, but maybe I'm misunderstanding.

src/components/SizeObserver/__snapshots__/SizeObserver.test.tsx.snap
```diff
@@ -0,0 +1,19 @@
...
+</span>
+`;
+
+exports[`<SizeObserver /> when its size changes twice but both changes match the same query renders at the corresponding size 1`] = `
+<span>
+  Other
```
chrismohr: ... and `Large` here?

src/components/SizeObserver/SizeObserver.md
chrismohr: @jessemartin The other examples use this to add data attributes that are consumed by our visual diff library.  

For example, on Box, 
```
```js { "props": { "data-description": "basic" } }
```

The filename for the visual diff will include "basic" as the name of the example.

PR #446, opened 2018-07-05
Remove dependency on @uifabric

unindented:
* Replace imports for `@uifabric/styling` and  `@uifabric/utilities` with `office-ui-fabric-react`.
* Bump version of `office-ui-fabric-react` to latest.


package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "normalize.css": "6.0.0",
-    "office-ui-fabric-react": "6.9.0",
+    "office-ui-fabric-react": "6.25.0",
```
chrismohr: I bumped this back to `6.25.0` since `6.25.1` and greater breaks our dropdown.  I plan to investigate further and either fix here or file an issue with fabric.

PR #449, opened 2018-07-06
chore: Upgrade all possible dev dependencies

unindented:
Leaving the following things out:
- `eslint`: Version 5.x is incompatible with current Airbnb config that we use.
- `ts-loader`: Versions 4.x rely on Webpack 4.x.
- `typescript`: Can't upgrade to 2.9.x as it complains about our implementation of `Omit`. The solution is have `Omit` use the new `Exclude`, but then all consumers of YamUI would have to be on 2.9.x too.
- `webpack`: Styleguidist still relies on Webpack 3.x.


src/components/Avatar/Avatar.styles.ts
```diff
@@ -10,20 +10,22 @@ export interface AvatarStyleProps {
...
+      initials,
+      image: {
+        borderRadius,
+      },
+      imageArea: {
+        borderRadius,
+      },
+    });
+  },
+);
```
unindented: I think this is prettier doing its thing, cause I didn't touch the files.

unindented:
Nah, got the same thing, where all screenshots differ.

PR #450, opened 2018-07-06
update the readme for release tags

chrismohr:
Because we typically squash-merge the release tags that we were pushing up were pointing to orphaned commits, rather than ones in master.


jessemartin:
👍

PR #451, opened 2018-07-06
add size limit for a demo component

chrismohr:
Example size-limit failure: https://travis-ci.org/Microsoft/YamUI/builds/401033546

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "react-test-renderer": "16.3.2",
     "resize-observer-polyfill": "1.5.0",
     "rimraf": "2.6.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -65,6 +65,7 @@
     "react-test-renderer": "16.3.2",
     "resize-observer-polyfill": "1.5.0",
     "rimraf": "2.6.1",
+    "size-limit": "^0.18.3",
```
unindented: We probably want an exact version here.

chrismohr: Right.  Thanks for catching.

unindented:
Good stuff. Minor comment.

PR #452, opened 2018-07-06
Upgrade dev dependencies

chrismohr:
reopening https://github.com/Microsoft/YamUI/pull/449


src/components/Avatar/Avatar.styles.ts
```diff
@@ -10,20 +10,22 @@ export interface AvatarStyleProps {
...
-    initials,
-    image: {
-      borderRadius,
-    },
-    imageArea: {
-      borderRadius,
-    },
-  });
-});
+export const getPersonaCoinStyles = memoizeFunction(
```
chrismohr: seems that the rules in .prettierrc aren't being respected by the git hook anymore

chrismohr: I've bumped prettier back down

PR #454, opened 2018-07-10
Create Tile component

unindented:
Added a `Tile` component that respects aspect ratio, which will become useful when rendering attachments.

## Pull request checklist

* [X] Component `README.md` file is up-to-date.
* [X] Component is unit tested.


src/components/Tile/Tile.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { theme } = classNameProps;
+
+  return mergeStyleSets({
+    root: {
+      border: `1px solid ${theme.semanticColors.bodyDivider}`,
+      selectors: {
+        '&.y-tile': {
+          position: 'relative',
+          width: '100%',
+          backgroundColor: theme.semanticColors.bodyDivider,
```
chrismohr: Is there a reason not to move these 3 up to the root?

unindented: Ah, no reason. I may have copied something I saw in `Text`.

src/components/Tile/Tile.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+        },
+        '&.y-tile.y-tile--ratio_21_9': {
+          paddingBottom: '42.85%',
+        },
+        '&.y-tile.y-tile--ratio_16_9': {
+          paddingBottom: '56.25%',
+        },
+        '&.y-tile.y-tile--ratio_4_3': {
+          paddingBottom: '75%',
```
chrismohr: If possible, I think it would be better to pass the aspectRatio to getClassNames, and set paddingBottom based on that, rather than adding the `y-tile--ratio...` class.

src/components/Tile/Tile.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+import { TileAspectRatio } from './Tile.types';
+import { memoizeFunction } from 'office-ui-fabric-react/lib/Utilities';
+import { mergeStyleSets, ITheme } from 'office-ui-fabric-react/lib/Styling';
+
+export interface TileClassNameProps {
+  aspectRatio: TileAspectRatio;
+  theme: ITheme;
+}
+
+const mapAspectRatioToPercent: Record<TileAspectRatio, string> = {
```
chrismohr: Nice use of `Record`!

src/components/Tile/Tile.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  [TileAspectRatio.TwentyOneNine]: '42.85%',
+};
+
+export const getClassNames = memoizeFunction((classNameProps: TileClassNameProps) => {
+  const { aspectRatio, theme } = classNameProps;
+
+  return mergeStyleSets({
+    root: {
+      backgroundColor: theme.semanticColors.bodyDivider,
+      border: `1px solid ${theme.semanticColors.bodyDivider}`,
```
swese44: Would we place anything besides an image as Tile content? Since this is a layout component, I would consider whether we want to enforce a border here or if that would be up to the application to place on the content.

Also note that the previous implementation used an inset 1px shadow overlay to darken the image's outer pixels rather than pushing the image in and placing the border outside. Not sure what the Designs call for, but looking at the "with text" example you probably want to do inset shadow instead of border.

unindented: For now, the only designs we have all have a border around the content, regardless of whether they're images or text.

True about the inset shadow, I'll fix that.

PR #456, opened 2018-07-10
Add Dark Theme

chrismohr:
- add dark theme
- see dark theme in dev by running `npm run start:dev:dark`

yamui_v8_8_0_dark


chrismohr:
You can also see dark themed components on a default styleguide, with `?theme=dark`.
dark_theme_default_styleguide

config/styleguide/context/redirectToDark.js
```diff
@@ -0,0 +1,11 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+/* eslint-disable */
```
unindented: It didn't like something you were doing?

config/styleguide/darkConfig.js
```diff
@@ -0,0 +1,24 @@
...
+      light: '#898989',
+      lightest: '#333333',
+      link: '#1978c8',
+      linkHover: '#f28a25',
+      border: '#B5B5B5',
+      name: '#7f9a44',
+      type: '#b77daa',
+      baseBackground: '#000',
+      codeBackground: '#0A0A0A',
+      sidebarBackground: '#0A0A0A',
```
unindented: My OCD forces me to bring up the fact that you're mixing lowercase and uppercase hex.

chrismohr: My ocd noticed it too and forgot to come back to it, thanks for catching it, I'll clean it up.

config/styleguide/context/redirectToDark.js
chrismohr: copy pasta from the action.js file.
But just checked, and it doesn't like document and window. 
```
[eslint] 'document' is not defined. (no-undef)
```

unindented: Ah, we probably need to tell `eslint` that we're in a browser environment:

```json
{
    "env": {
        "browser": true
    }
}
```

config/styleguide/config.js
```diff
@@ -7,6 +7,15 @@ const pkg = require('../../package.json');
 
 const componentsWithoutPathLine = ['Icons'];
 
+const redirectToDarkScript = `const redirectToDark = () => {
+  if (document.title.match(/DARK/) && !window.location.href.match(/theme=dark/)) {
```
unindented: Why also rely on `document.title`?

unindented: Haha funky, but ok.

chrismohr: :-) Curious which part you thought was funky?  Having the code in a string? (Is there a better way?) Or the redirect logic?

unindented: The code in a string, and checking the document title + lack of parameter in the URL to decide whether to redirect. I don't have better alternatives though, so my comment wasn't very useful.

chrismohr: I like to hear comments like this even when you don't have suggestions to improve it.  As the code writer, it's not always obvious which code will be funky to others, so I think it's useful.

I was able to make an incremental improvement to the logic.  I took the title bit out, since it's now simpler to just add the script to the darkConfig file.

config/styleguide-visual/config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
   },
   launchOptions: {
     args: ['--no-sandbox', '--disable-setuid-sandbox', '--font-render-hinting medium'],
-    slowMo: 10,
+    slowMo: 15,
```
chrismohr: Bumped this up, since I saw a couple flakey visual diff failures yesterday.

PR #466, opened 2018-07-18
Feature(YamUI) Adding HighContrastSelector component, stylesheet, test and snapshot.

Logrologist:
Adding HighContrastSelector component and corresponding test, stylesheet and snapshot.
- This component will utilize a CSS @media query with -ms-high-contrast to determine which version of a particular component to display.
- Adding component to YamUI markdown view
## Pull request checklist


src/components/HighContrastSelector/HighContrastSelector.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+    props = {
+      regular: 'regular',
+      highContrast: 'highContrast',
+    };
+
+    component = shallow(<HighContrastSelector {...props} />);
+  });
+
+  describe('render with props', () => {
+    it('hides the highContrast element', () => {
```
BrendanMcK: I'd just say 'renders as expected' here - unfortunately we can't actually determine whether it's hiding/showing any specific component by looking at the snapshot alone.

src/components/HighContrastSelector/HighContrastSelector.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default class HighContrastSelector extends React.Component<HighContrastSelectorProps> {
+  public render() {
+    const { regular, highContrast } = this.props;
+
+    return (
+      <>
+        <div className="y-high-contrast-selector--regular">{regular}</div>
+        <div className="y-high-contrast-selector--highContrast">{highContrast}</div>
+      </>
```
unindented: Can you think of a way of not rendering both on the page, in case they're costly to render?

I was thinking something like:

```
<>
  <div className="y-high-contrast-selector__canary" ref={this.setRef} />
  {!this.state.isHighContrast && <div className="y-high-contrast-selector--regular">{regular}</div>}
  {this.state.isHighContrast && <div className="y-high-contrast-selector--highContrast">{highContrast}</div>}
</>
```

Then when `setRef` gets called, we check its `display` or some other attribute we set with CSS, and call `this.setState({ isHighContrast: true/false })`

src/components/HighContrastSelector/HighContrastSelector.tsx
```diff
@@ -0,0 +1,22 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import * as React from 'react';
+
+import 'HighContrastSelector.css';
```
chrismohr: Can you please replace this with a getClassNames function? We've been working towards using merge-styles for all components.

The Box component will have an example to follow. 

In future, if you generate the component with `npm run create:component` it'll create the files you need.

chrismohr: You can do the media queries like this:
https://www.npmjs.com/package/@uifabric/merge-styles#media-queries

chrismohr: It looks like you can also import the selector from fabric if you want:
```import { HighContrastSelector } from 'office-ui-fabric-react/lib/Styling';```

src/components/HighContrastSelector/HighContrastSelector.tsx
unindented: KISS then.

src/components/HighContrastSelector/HighContrastSelector.tsx
Logrologist: Yeah. I've already been working my way through this after having tried out: npm run create:component (just to see what that did vs what i made from scratch). I hadn't realized that we had a working React-flavored way to handle media queries, so I was sticking with the CSS. I'll add the .styles stuff back in.

src/components/HighContrastSelector/HighContrastSelector.test.tsx
Logrologist: Fixed it.

PR #469, opened 2018-07-19
make fabric a peer dependency

chrismohr:
- hoping to mitigate version conflicts for transitive dependencies in projects that are also importing office-ui-fabric-react


package.json
```diff
@@ -77,6 +77,7 @@
     "yo": "2.0.3"
   },
   "peerDependencies": {
+    "office-ui-fabric-react": "6.25.0",
```
chrismohr: Using an exact version number to ensure quality.  If we allow greater version numbers there's a risk of undetected regressions making it into the product.

PR #470, opened 2018-07-19
allow SuggestionList to use inverted Theme

chrismohr:
Before:
before

After:
after


unindented:
What's the meaning of inverted (vs dark vs HC)?

chrismohr:
@unindented  Good question. Here's my take:
- "inverted" is an attribute of a fabric theme, I'd say that it's appropriate to set this to true when the color named "white" in the color palette is darker than the color named "black".
- "dark" is the name of a fabric theme that we (developers) created for testing theming in Yammer. AFAIK there are no plans to offer this to end users.
- "HC" (High Contrast mode) is browser setting.  When active, the browser applies a set of styles to increase the visibility of content on the page. In YamUI , we detect this mode, and apply styles to enhance/patch bugs.

I haven't yet tested for interactions between HC and our "dark theme".

PR #473, opened 2018-07-20
add borderColor prop to Box

chrismohr:
- also fix bug where setting backgroundColor on box was covering the border


unindented:
Shouldn't `Box` always use the `borderColor` from the theme? What are use cases where we need to customize the border color?

chrismohr:
@unindented We only have one concrete case right now.  The focused state of the Editor (Publisher) is blue.  A second case might be the active state of the Navigation buttons at the top left.  They don't currently use Box, but I think they might be a bit simpler if they did.

I'm definitely wary of adding to many specific style props to components, but in this case was fine with it since the component is simple, and the prop is closely related to the primary function of the component, which is being a border. :-)

PR #476, opened 2018-07-20
update image files, but see no visual differences

chrismohr:
I'm not quite sure what's going on here. 

- visual tests pass
- I run `npm run test:visual:approve`
- these files become dirty

On commit, it shows them as `rewrite` of 94-99% 

Any ideas?


chrismohr:
There's a binary diff, but no visual diff. 
bin-diff

unindented:
Uuuuh. Then you're right, `react-styleguidist-visual` should only approve files for which there's a `.diff.png`.

PR #483, opened 2018-08-03
fix(docker visual tests): prevent styleguideist from picking up stale dist assets

jessemartin:
If npm build is run, then an breaking change (such as a new TS version) is introduced, docker visual
tests will fail unless npm run clean:dist is manually run. This change removes the need to do this,
by running clean:dist as part of the docker test commands.


jessemartin:
@chrismohr

PR #501, opened 2018-08-16
Create ActionLinkList component to enforce ActionLink padding in sidebar

### LINKED ISSUE #498: Create ActionLinkList component to enforce padding of action links in right rail
since we removed its padding here:
https://github.com/Microsoft/YamUI/pull/497

swese44:
Renders a vertical list of ActionLink components for sidebar widgets. Fixes #498, i.e. replaces the padding we removed to reuse ActionLink outside of the right rail.

## Pull request checklist


src/components/ActionLinkList/ActionLinkList.styles.ts
```diff
@@ -0,0 +1,15 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import { mergeStyleSets } from 'office-ui-fabric-react/lib/Styling';
+import { memoizeFunction } from 'office-ui-fabric-react/lib/Utilities';
+
+export const getClassNames = memoizeFunction(() => {
+  return mergeStyleSets({
+    root: {
+      selectors: {
+        '.y-actionLink': {
```
chrismohr: Curious what you think of this idea... it's a bit different than what we've been doing, but where I'm at right now, I'd prefer to avoid styling our own class names, since I think it's a bit more brittle than a styles prop, or using mergeStyles.  It's not too big of a deal, but if we rename the actionLink class we wouldn't find out that something is broken here until the visual regression tests.

I'd like it we could start a pattern of using a styles prop for stuff like this like fabric has. I think it would be better than adding a mergeStyle here, because it would lead to fewer class/styles in total.

It might be too much for this PR, but thought I'd bring it up to get feedback on the idea. 

So the ideal in my mind right now would be if we could do something like: 

```
<LayoutListItem styles={{root: { padding: '0.2rem 0 0.6rem' } }} ...
```
(with the styles coming from a function in the .styles file)

chrismohr:
I'm probably just being weird here, but the icons look a bit crowded to me, was there a reference image you were working against?

src/components/ActionLinkList/ActionLinkList.styles.ts
swese44: i do like this idea within YamUI, but i'd prefer not to expose this externally unless we run into some crazy use case (since YamUI tries to alleviate this type of style overrides altogether). let's continue to think about this

src/components/ActionLinkList/ActionLinkList.md
```diff
@@ -0,0 +1,21 @@
...
+  <ActionLink text="Download" icon={Download} ariaLabel="Download this file" onClick={action('ActionLink clicked')} />,
+  <ActionLink text="Upload a new version of this file" icon={Upload} onClick={action('ActionLink clicked')} />,
+  <ActionLink text="Group Info" icon={Group} ariaLabel="Go to group info page" href="https://www.yammer.com" newWindow={true} />,
+  <ActionLink text="SharePoint Document Library website" icon={SharePoint} href="https://www.yammer.com" newWindow={true} />,
+];
+
+<div style={{ maxWidth: '300px' }}>
+  <ActionLinkList items={items} />
+</div>
+```
```
chrismohr: I could take it or leave it, but might be nice to have one example wrap to a next line?

swese44:
@chrismohr mockups are here:

this adds the same padding back that Hanna had removed from ActionLink, height still matches 28px from the mockups

src/components/ActionLinkList/ActionLinkList.md
swese44: cool, updated

PR #504, opened 2018-08-17
feat(tile): add eight by nine aspect ratio

jessemartin:
Adds an eight by nine tile aspect ratio, which will be used in the image collage component in modern
client.


jessemartin:
@chrismohr @swese44 @richardtagger Please review.

PR #507, opened 2018-08-21
Remove global CSS resets and base styling

### LINKED ISSUE #1: Fabric bug loses certain typography styles when provided with a theme
When providing a window.FabricConfig object with font styles, Fabric has a bug where a shallow-assign happens instead of merging styles. For example, when setting -webkit-font-smoothing: auto we lose font-family and other rules.

I'm temporarily adding the extra 3 @font-face rules in our CSS to compensate, which we can remove once this bug is fixed in Fabric:
https://github.com/OfficeDev/office-ui-fabric-react/issues/2013

### LINKED ISSUE #488: Base Component with Styles
When embedded in another page, our base css conflicts with the hosting page.

We should use merge-styles to add all the styles we need to this component, so that the consumer no longer needs to import our css file, and all styles are only applied where they're supposed to be.

swese44:
Fixes #488. Fixes #1.

This change was primarily driven to support SP web parts which cannot have our global resets and baseline styles. Since MC is entirely YamUI (Fabric) we no longer need to ensure baseline styles for all elements are set.

This is a breaking change which no longer includes base element styling. This will be a new major version which will not be compatible with YamJS.

THIS IS A BIG PR. IT'S PROBABLY EASIEST TO REVIEW EACH COMMIT SEPARATELY.

Changes:
- Removed normalize.css, our reset styles, and baseline element styles
- Created Paragraph, Strong and Emphasis component to be used instead of their HTML tags (can't style the tags globally now and expect SP web parts to render them reliably)
- Removed the Fabric wrapper component export, provided AppContainer which nests a Fabric component but also includes a couple global html tag styles and some app-root scoped styles.
- Fixed a long-standing issue where Fabric wasn't including some base font styles like font-family due to it not deeply-merging our theme styles. Also removed our hacky solution which linked to the fonts directly. This should also resolve any blockers preventing our app from automatically rendering the correct locale-specific font family based on lang attr.
- Updated visual diffs. Many diffs had 1 or 2 pixels difference, mostly in aliasing SVG icons. I don't see anything which should actually be a blocker.

Which things which were not removed, and why:
- box-sizing: border-box at root html element. This is necessary because things like Picker are rendered outside of our AppContainer component. This has not affected SP web parts.
- Font size at root html element set to 62.5%. This is required to support our font scaling strategy so our fonts can scale up with the user's custom base font size. SP web parts remain unaffected, Fabric text styles all use px values. Eventually we'll conform with Fabric under Fluent and our custom strategy will go away.
- High-contrast classnames and `a` tag styles in our baseline CSS. These are needed by MC at least for the time being, and SP web parts can simply not include YamUI's base CSS file.

What this means for MC:
- Just need to replace `<Fabric>` components with `<AppContainer>`

What this means for SP web parts:
- Just stop importing YamUI's baseline CSS file

## Pull request checklist


richardtagger:
Thanks for working on this @swese44 !

PR #509, opened 2018-08-21
Remove margin from body tag, increase Styleguidist font size

swese44:
The only issue i see in MC is margin on the body tag, this fixes it. Also increases Styleguidist font size -- it's importing the base 10px size from the html tag, since it no longer gets the 1.5rem base font size.

## Pull request checklist


src/css/base.css
```diff
@@ -1,6 +1,12 @@
 /*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
 @import 'src/css/variables/colors.css';
 
+html,
+body {
+  margin: 0;
+  padding: 0;
+}
+
```
chrismohr: Is there a reason not to put these on the AppContainer?
https://github.com/Microsoft/YamUI/blob/master/src/components/AppContainer/AppContainer.styles.ts#L21

swese44: i guess just to avoid more global CSS for SP, but this wouldn't affect their page. should i move the rules over to AppContainer?

chrismohr: Oh.. I see.  If it's specifically for the modern web client.  Maybe best would be to add the css there?

swese44: updated

PR #511, opened 2018-08-22
Add HorizontalActionList Component

chrismohr:
Mocks: https://app.zeplin.io/project/5907d6451d2b06c6b36b2852/screen/5b7b17c4336cfa21f21efe87

Introduces a variation to ActionLink called "compact", which makes the link bold, has a smaller icon, and smaller gutter.


swese44:
what if ActionLink accepted additional props to support this? then ActionLinkList would just need a `direction` prop

chrismohr:
@swese44 I like the idea of ActionLink accepting additional props to achieve the correct look, there are only two things that are different. The iconSize and gutterSize are smaller. I see couple of options to handle this.  Right now I like adding a `compact: boolean` prop for this, we could also do a `size:  'normal' | 'small'`, or just expose both `iconSize` and `gutterSize`.  I think I like one of the first two options to limit options.

It would look like this:
```
    const gutterSize = compact ? GutterSize.XSMALL : GutterSize.SMALL;
    const iconSize = compact ? IconSize.MEDIUM : IconSize.MEDIUM;
```

2. The next decision might be around what to do with the count. We could add it to actionLink as an `unlinkedSuffix` or `unlinkedText`, but I think I like leaving it out of action link, and bringing it up a level, since it's layout (6px spacing) seems pretty specific to this case.  

3. Then for ActionLinkList, I'm interested to hear how you imagined it working with the overflow menu.  Would you pass in the overflow menu as an item? Or would you have it the menuButton  in a column outside the ActionLinkList? Or would you add props/logic to ActionLink List to separate the items into the visible and overflow sets? Something else?

src/components/ActionLink/ActionLink.types.ts
```diff
@@ -16,7 +16,12 @@ export interface BaseActionLinkProps extends BaseComponentProps {
   /**
    * The visible text.
    */
-  text: string;
+  text: string | React.ReactNode;
```
swese44: Should this remain only string? it looks like you solved this with only string and we probably don't want to allow arbitrary markup here?

src/components/MessageActionsList/MessageActionsList.tsx
```diff
@@ -0,0 +1,69 @@
...
+import ActionLink from '../ActionLink';
+import { FixedGridRow, FixedGridColumn } from '../FixedGrid';
+import Block, { TextSize } from '../Block';
+import { IconSize } from '../Icon';
+
+/**
+ * A set of links that dispaly in a horizontal list below a message post.  When a maxVisibleItemCount
+ * prop is provided, items that exceed this value will show in a overflow menu.
+ */
+export class MessageActionsList extends React.Component<MessageActionsListProps & CustomizableComponentProps> {
```
swese44: it would be nice to name this something more generic so can be associated with a content item rather than Message. i'd expect to see yammer-web-components owning a `<MessageActionsList>` component which feeds the correct message-specific actions to this reusable UI component. maybe this could just be HorizontalActionList, or CompactActionList, ContentActionList... i don't think i'm coming up with any winners though

src/components/MessageActionsList/MessageActionsList.types.ts
```diff
@@ -0,0 +1,30 @@
...
+   */
+  overflowMenuAriaLabel: string;
+
+  /**
+   * The maximum number of items to display in the list.  The remaining will show in the menuButton.
+   */
+  maxVisibleItemCount?: number;
+}
+
+export type MessageActionsListItem = {
```
swese44: seems like we're duplicating ActionLink props, but losing the nuance of it's compound prop types (link or clickable). should we remove this interface and set line 9 to `items: ActionItemProps[]`?

src/components/MessageActionsList/MessageActionsList.tsx
chrismohr: I agree. @cainhimself do you have any ideas or opinions here?

src/components/MessageActionsList/MessageActionsList.types.ts
chrismohr: Yeah, I agree this could be better. There are two things that lead me to create the new interface here. 
1. ActionLink has an option `ariaLabel`, `menuButton` requires it.
2. Currently unlinkedText is rendered by the list component, not the ActionLink. 

I'll try to figure something out.

swese44: so menuButton options have to be onclick handlers, not hyperlinks?

swese44: your call on this, whether it makes sense to extend ActionItemProps[] or not

src/components/AppContainer/AppContainer.tsx
```diff
@@ -15,7 +15,7 @@ export default class AppContainer extends React.Component<NestableBaseComponentP
     const { children, className, theme = defaultTheme } = this.props;
     const classNames = getClassNames({ theme });
     return (
-      <Fabric>
+      <Fabric theme={theme}>
```
swese44: might need to double check this when you merge from master as Donnie's PR has been merged

chrismohr: Yeah, I see it's conflicting.  I'll resolve it.

swese44:
is there anything blocking this merge so we can start the MC import?

chrismohr:
@swese44, please feel free to merge this if it unblocks you.

@cainhimself is looking for changes to the menuButton spacing, but I'm happy to do this in a new PR next week.

swese44:
cool thanks

PR #512, opened 2018-08-22
[spike] render null if FixedGridColumn is empty

chrismohr:
Not sure if we want to do this, but I was curious to see what regressions there would be.
----

Helps when you need to avoid the extra whitespace for an optional column.

Instead of:
```		
   const optionalSecondThing = optionalSecondThing ? (
      <FixedGridColumn fixed={true}>
           { optionalSecondThing }
      </FixedGridColumn>
    ) : null;

  return (
  <FixedGridRow>
    <FixedGridColumn>{firstThing}</FixedGridColumn>
    {optionalSecondThing}
  </FixedGridRow>)
```

You could do:
Instead of:
```		
  return (
  <FixedGridRow>
    <FixedGridColumn>{firstThing}</FixedGridColumn>
    <FixedGridColumn>{optionalSecondThing}</FixedGridColumn>
  </FixedGridRow>)
```


swese44:
the problem is in many cases we still need to render the empty column even if there isn't content. so implicitly not rendering the column would introduce some bugs in our apps, and i think most people would be surprised if a column never appeared even though they are clearly rendering a `<FixedGridColumn>` component

chrismohr:
I agree that empty columns could be surprising.

PR #515, opened 2018-08-27
Remove global fabric theme load

dkarns:
### DO NOT MERGE

Putting this PR up to present visual diffs to devs for analysis/approval

## OVERVIEW

This PR removes the global fabric theme load from YamUI to avoid overriding existing global themes by default.

Removing global theme produced a number of visual diff failures -- partially due to using an older version of Fabric which didn't support theming across all components.

I bumped Fabric to the most recent release (that YamUI could successfully build) which reduced the number of visual diff failures, but there remain a dozen or so.

These may be valid / good changes to YamUI; submitting this PR so that can be reviewed.


swese44:
how were you able to check in the .new.png and .diff.png visual images? they are only used as reference during the visual diff tests and are blocked by .gitignore...

dkarns:
I used `git add -f /path/to/blah.diff.png` for each diff with failures.

I checked them in (with a big ol' DO NOT MERGE) so you all could look at the diffs more easily.

swese44:
`npm run test:visual:approve` replaces the previous diff image with the new one. if an image file has changed then Github will show a few diffing options so you can slide or fade between images and see the differences pixel-by-pixel

src/components/AppContainer/AppContainer.tsx
```diff
@@ -15,7 +15,7 @@ export default class AppContainer extends React.Component<NestableBaseComponentP
     const { children, className, theme = defaultTheme } = this.props;
```
swese44: hey @chrismohr, defaulting to defaultTheme in the AppContainer seems correct. but all of our other components also have this `theme = this.props.theme || defaultTheme`. is this correct, or will every component now default to the default YamUI theme instead of what was passed into the app root?

chrismohr: @swese44  The reason we do `theme = this.props.theme || defaultTheme` is to reassure typescript that the theme prop is not null. When the component is decorated with @customizable, it will always have a `theme` prop. I see how this is confusing though.  Any ideas for a better solution?

swese44: cool thanks. i'm going to push up a commit which updates AppContainer to properly set the incoming theme, and updates ExampleWrapper to only set dark theme if provided (and let AppWrapper do the defaulting)

swese44:
looks like a few styling issues from the Fabric update
- dropdown menu items need padding adjusted
- dropdown chevron icon needs position adjusted
- Dropdown_with_templates_dividers_headers_click_desktop.png, Dropdown_with_templates_dividers_headers_keypress_arrowdown_desktop.png lost their open state
- Dropdown_with_templates_dividers_headers_keypress_enter_desktop.png screenshotted the wrong box area... weird
- Picker_tag_picker_click_desktop.png selected item pills stack vertically instead of horizontally
- TextArea_auto_adjust_height_desktop.png TextArea no longer grows automatically
- TextArea_error_desktop.png is the wrong height. i think the removed line-height rule needs to be put back, it's important for multiple lines of text

i can't imagine managing a UI library without visual diffs, they've saved our bacon so many times

src/components/Picker/Picker.styles.ts
```diff
@@ -25,5 +36,35 @@ export const getClassNames = memoizeFunction((classNameProps: PickerClassNamePro
...
+        '.ms-Suggestions-item[class*="suggestionsItemIsSuggested"]': {
+          backgroundColor: theme.palette.neutralLight,
+          selectors: {
+            ':hover': {
+              backgroundColor: theme.palette.neutralTertiaryAlt,
+            },
+          },
+        },
+      },
+    },
```
dkarns: I think there should be a way to do this cleaner by using mergeStyles's parent/child referencing features:
https://github.com/OfficeDev/office-ui-fabric-react/wiki/mergeStyles#referencing-child-elements-within-the-mergestylesets-scope

I gave it a first shot and the selectors stopped working. I'll try to revisit later.

swese44: i'm not thrilled with it either, but it don't see an obvious better way without a refactor, or the upstream Picker getting converted to mergeStyles first. let's leave this one as-is because we'll soon be removing all of our Fabric style overrides anyway

src/components/Picker/Picker.styles.ts
```diff
@@ -25,5 +36,35 @@ export const getClassNames = memoizeFunction((classNameProps: PickerClassNamePro
...
+    suggestionsList: {
+      selectors: {
+        '.ms-Suggestions-item': {
+          selectors: {
+            ':hover': {
+              backgroundColor: theme.palette.neutralLighter,
+            },
+          },
+        },
+        '[aria-selected="true"]>.ms-Suggestions-item': {
```
dkarns: @swese44 Whoops. I just pushed this selector change not considering how it would affect the snapshot changes you just pushed.

Decided to use `aria-selected` attribute to determine whether we're looking at the currently-selected picker item rather than the classname in order to be consistent with my other selector in Dropdown, and because that seems a more safe selection assumption than the classname.

Do you prefer this? If so, I'll update the snaps myself. If not, or if you don't care one way or the other, I'll just revert.

dkarns: Meh. I'm just going to revert.

src/components/TextField/__snapshots__/TextField.test.tsx.snap
```diff
@@ -13,7 +13,7 @@ exports[`<TextField /> with all options matches its snapshot 1`] = `
   maxLength={100}
   multiline={false}
   onBlur={[MockFunction onBlur]}
-  onChange={[Function]}
```
dkarns: Hm... isn't `onChanged` the one that's deprecated in Fabric's `TextField`?

https://github.com/OfficeDev/office-ui-fabric-react/pull/5764

https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/TextField/TextField.types.ts#L167

dkarns: NEVERMIND... this diff is a revert of @swese44 's attempt to move us over to `onChange`.

PR #522, opened 2018-09-05
feat(focus): Add focus outlines to NavigationLink and Clickable

BrendanMcK:
Add focusOutlines style helper that returns styles that NaviagionLink
and Clickable merge into their own styles.

Initial version is provididing just a basic box-shadow-based outline,
only showing it when needed. We avoid using Office UI Fabric's focus
styles, as they assume block elements, and don't work well on non-block
links that wrap. Can consider switching to their style in future if they
resolve that issue (https://github.com/OfficeDev/office-ui-fabric-react/issues/4883)

Later commit will also add option to use a overlay focus style, which
is needed for links and buttons that contain images or other content
that covers their surface area competely - main use case there is for
eg. images in a gif picker, or image attachments.


swese44:
looks like some hover or focus styles have changed in the visual diffs

PR #529, opened 2018-09-11
feat(Pivot): Support the ability to specify Pivot item sizes.

benlakey:
Fabric supports the ability to customize the size of Pivot items.
Previously we were not supporting that. This change allows pivot items
in YamUI to have the size prop that Fabric accepts.

## Pull request checklist


src/components/Pivot/Pivot.tsx
```diff
@@ -42,6 +48,7 @@ export default class Pivot extends React.Component<PivotProps> {
         selectedKey={this.props.selectedKey}
         onLinkClick={this.onLinkClick}
         headersOnly={true}
+        linkSize={this.props.linkSize ? this.props.linkSize : PivotLinkSize.normal}
```
BrendanMcK: We luck out here in that PivotLinkSize.normal is 0 - so if this prop is undefined *or* normal - we'll use normal. Explicit undefined check might make intent more clear here?

benlakey: Are you thinking it would be safer to just not pass the prop at all when its left unspecified?

BrendanMcK: Looks like we can just pass this right though.

While Fabric internally isn't comparing against unknown, it's also defaulting to normal size, and only adding in its extra classes for large if the linkSize is explicitly === to PivotLinkSize.large

benlakey: Updated

PR #533, opened 2018-09-20
refactor(YamUI): Modify indentation of ChoiceGroup component

eoh:
Modify indentation of ChoiceGroup component to more closely match polls on Yammer.com (https://yammer.visualstudio.com/frontend/_workitems/edit/19062)

Currently on Modern Client:
screen shot 2018-09-19 at 5 14 56 pm

With this PR change:
screen shot 2018-09-19 at 5 11 41 pm


BrendanMcK:
My 2c is that this works *better* as-is - where the checkbox is to the left of an aligned paragraph. I think that this is the standard for checkboxes - and it's yamjs that's the wonky one here...

eoh:
@BrendanMcK I agree 1000%: https://www.yammer.com/microsoft.com/#/Threads/show?threadId=1158532280

Issue #527, opened 2018-09-10
HorizontalActionList has a type in folder name


allobast:
`HorizontalActionList` component has a type in the folder name - it is HorizontalAction`s`List and it causes confusion when importing.

PR #535, opened 2018-10-28
Fix typo in HorizontalActionsList folder

kleinrein:
Removed an "s" in the folder HorizontalActionList to fix issue #527.


swese44:
hey thanks @kleinrein. unfortunately we've rolled back YamUI to an older version and won't be actively maintaining new features other than what's necessary to support our applications still using YamUI. if the library is useful i'd recommend forking it from v9.x, see the updated project README. thanks!

PR #538, opened 2018-11-08
Roll back to 8.2 release, keeping newer Icons and Illustrations

swese44:
I would recommend reviewing individual commits as the first revert commit is pretty large


tonyja:
Individual commit reviews:

---
Revert last 141 commits, back to 8.2.0 release - ae528c3
It looks like the top level README andicon and illustration README may want more of the master changes kept in them.  I am not sure, but I think some of the reverts show actually be kept instead.  Details:
- Did you intend to remove the illustrations README.md file?  I would think not:
https://github.com/Microsoft/YamUI/pull/538/files#diff-b0bfee02124a3cccdd0937247cb744a6
- Is the icons README text correct with these reverts?  Does the state in this PR match the guidance you want to give for the 8.2.0 version of YamUI or does it need more editing?  I could not tell if these instructions were valid or if the reverted changes were something we wanted on this new 8.2.0 based master or not:
https://github.com/Microsoft/YamUI/pull/538/files#diff-206223ec855878fad82fe1a087abf521
- I would also review the core README to ensure you want the adding icons, pushing versions, and size limit changes to be reverted (are they good edits to keep with the new master or are they intentionally reverted to match the 8.2.0 state) and also to look at the "Roadmap" part above the "Contributing" part that contains "pre-release state" text, we may want that section removed now in light of the new maintenance section:
https://github.com/Microsoft/YamUI/pull/538/files#diff-04c6e90faac2675aa89e2176d2eec7d8L84
---
These two readme changes look fine to me. (with minor commit message comment below)
- Add maintenance mode note in the README - ffd6523
- Keep codeowners -1a18e0a 

---
Keep newer Icons and Illustrations - 8c49222
I compared the 8.2.0 version of the visual diff and the current master version and your version and it looks like all the missing Icon_library_desktop SVG entries from 8.2.0 are re-added and present in your PR so it should match the master icons:

ICONS
8.2.0 is missing only the FileBug.svg icon- https://github.com/Microsoft/YamUI/blob/3200fc7fcb756a61d8c4612f4a8046de6c4db1d7/config/styleguide-visual/reference/Icons_library_desktop.png
master and your PR have the same icons - 
https://github.com/Microsoft/YamUI/blob/master/config/styleguide-visual/reference/Icons_library_desktop.png

ILLUSTRATIONS
8.2.0 is missing 36 illustrations (the praise ones) - 
https://github.com/Microsoft/YamUI/blob/3200fc7fcb756a61d8c4612f4a8046de6c4db1d7/config/styleguide-visual/reference/Illustrations_library_desktop.png
master and your PR appear to have the same full set -
https://github.com/Microsoft/YamUI/blob/master/config/styleguide-visual/reference/Illustrations_library_desktop.png

---
(Minor note: Minor comment about codeowners sha 1a18e0a, the commit message looks like the wrong revert commit is mentioned, but this is minor and does not need to be amended. i.e. "This reverts commit 2d090fc ." but shouldn't it mention mention "Restore: Add Jesse and Brendan as approvers - 581e676 " or the "revert the revert commit sha ae528c3 )

PR #536, opened 2018-11-07
New QuestionMessage icon

tonyja:
*New QuestionMessage icon*
Adding the following QuestionMessage icon to support the new Question and Answer features:
screen shot 2018-11-07 at 2 44 52 pm

## Pull request checklist

* [x] Component `README.md` file is up-to-date.
* [x] Component is unit tested.

PR #548, opened 2018-11-09
Release version 10.0.0 (a revert to 8.2.0 with some additional icons and cleanup)

tonyja:
*Release a new major version for the continuing maintenance mode YamJS phase*

- See the following PRs for details on how the old 9.4.0 branch was reverted back to 8.2.0 (in order to be compatible with current YamJS) and then had some icons and cleanup code and dependencies updates added: #538 (revert + icons + tidying) and #547 (update lodash)

## Pull request checklist


tonyja:
- Also I am not sure what @swese44 would recommend for release versioning but in order to avoid confusion with 8.2.0 and 9.x.x and to highlight the significant change I created a PR for releasing master with `npm version v.10.0.0 -m "Release version %s"` that is on a master branch (already including PRs #538 and #547 ) and that can go out to YamJS as it is is in master now with the additional PR #536 QuestionMessage icon being a v10.0.1 after that pulled into YamJS separately.
- If it makes sense instead, we can include those two PRs now in master and also the PR #536 branch merged to master and rebased into here as well and go out to YamJS together in 10.0.0

I will rely on @swese44 and @chrismohr and others to advise on the best release version plan.

tonyja:
Thanks for the merge and release adjustment too.  When this is merged I should be able to rev the YamJS from 8.2.0 to 10.0.0 and use the QuestionMessage icon there in my work next week.

PR #555, opened 2019-02-07
feat(button): add newWindow prop

jessemartin:
Adds a `newWindow` prop to `Button`, so that link buttons can open with target="_blank".

## Pull request checklist


src/components/Button/Button.md
```diff
@@ -533,6 +533,12 @@ Link button:
 <Button text="I am a link" href="https://www.yammer.com" />
 ```
 
+Link button with newWindow:
+
+```js { "props": { "data-description": "link" } }
```
swese44: this needs a `different data-description` from the previous link, as-is the visual dff for this one overwrote the last one. (you'll also need to approve and commit the new diff image via Docker)

swese44:
looks good, just the one comment

src/components/Button/Button.md
jessemartin: Cool, will update!

